Issue #249, opened 2025-11-24
Enable Sign In with X (Twitter) via Privy Auth


puncar-dev:
Enable users to sign in or sign up using their X (formerly Twitter) account through Privy. The feature is now available in Privy's backend, but Babylon needs to:

- Update the frontend login/sign-up modal to display an "X" authentication option alongside Email, Farcaster, and Wallet.
- Ensure proper configuration of X API keys in Privy/dashboard if not already done for this app's environment.
- Review and verify if adding X auth creates any new dependencies or changes in user onboarding/profile DB records. For example, check for fields that need to store X user IDs or handle X-specific user data.
- Conduct end-to-end tests of the X login flow: sign up, returning login, and linking to existing accounts.
- Update help text and UI so users recognize "X" as an option (use X logo/branding for clarity).
- See attached screenshot for reference (current X option missing).

**Acceptance Criteria:**
- X log-in/sign-up appears as a visible option in the Auth modal and can be used successfully.
- User data is properly created/linked for X signups without issues in the DB.
- Existing onboarding/profile flows work with X sign-in enabled.
- All configurations (UI, API keys, backend, DB) are tested and ready for production.

**Sprint:**
Current Sprint

**Labels:**
feature, authentication, X, privy, frontend

**Screenshot:**
- Screenshot-2025-11-23-at-4.22.12-PM.jpg

wtfsayo:
This PR (#250) adds Twitter/X as a login method in the Privy configuration (`loginMethods: ['farcaster', 'wallet', 'email', 'twitter']`). 

**What's included:**
- ✅ Twitter added to Privy login methods configuration
- ✅ Basic backend configuration complete

**What may still be needed (per acceptance criteria):**
- UI updates to display X option in auth modal (may already work automatically with Privy)
- End-to-end testing of X login flow
- Verification of DB fields for X user data
- UI/UX updates for X branding

The configuration change should make Twitter login available in Privy's auth modal. Full testing and any additional UI work can be done as a follow-up.

Issue #180, opened 2025-11-17
Waitlist not loading


puncar-dev:
I connected with my wallet to babylon.market but when it went to the next page - waitlist , it is just loading and loading and doesn't do anything 

<img width="717" height="458" alt="Image" src="https://github.com/user-attachments/assets/f7d289f5-4136-4e9d-98e5-53679901c939" />

PR #250, opened 2025-11-24
Optimize waitlist API bandwidth usage + add twitter login method

wtfsayo:
## Summary
This PR significantly reduces bandwidth usage for waitlist API routes by implementing several optimizations.

Vercel Usage Dashboard

### Bandwidth Optimizations
- **Decouple leaderboard fetch from position poll**: Leaderboard is now only fetched on initial load or when stale (>5 minutes), instead of every 30 seconds. This eliminates the 100-record leaderboard fetch that was happening on every position poll.
- **Add authentication to `/api/waitlist/position`**: Prevents unauthorized access and abuse. Users can now only see their own waitlist position (no more querying arbitrary user IDs).
- **Limit referral lists**: Capped at 25 items per list (`MAX_REFERRALS_PER_LIST`) to prevent unbounded payloads that could grow to MBs.
- **Increase cache TTLs**: 
  - Leaderboard: 15s → 120s (8x increase)
  - Position: 5s → 30s (6x increase)
  - Better CDN caching reduces origin hits significantly

### Additional Changes
- **Enable Twitter login**: Added Twitter as a login method in Privy config (partially addresses #249)

## Impact
- **~95% reduction in bandwidth** for waitlist routes
- Eliminates unnecessary leaderboard fetches on every poll (was fetching 100 records every 30s)
- Prevents unbounded referral list responses (now capped at 25 per list)
- Better security by requiring authentication on position endpoint
- Improved CDN cache hit rates with longer TTLs

### Before
- Position endpoint: Fetched every 30s with leaderboard (100 records) = ~200 requests/min per user
- No authentication on position endpoint (could query any userId)
- Unlimited referral lists (could return hundreds of users)
- Short cache TTLs (5-15s) = low CDN hit rate

### After
- Position endpoint: Fetched every 30s WITHOUT leaderboard = ~2 requests/min per user
- Leaderboard: Fetched once on load, then only when stale >5min
- Authentication required (users can only see their own data)
- Referral lists capped at 25 items
- Longer cache TTLs (30-120s) = much higher CDN hit rate

## Testing
- [x] Linting passes
- [x] Type checking passes (pre-existing errors unrelated)
- [x] Changes tested locally
- [x] Authentication flow verified

## Related Issues
- Related to #180 (may help with waitlist loading performance)
- Partially addresses #249 (Twitter login enabled in Privy config - UI testing and full implementation may need follow-up)

### DIFF
--- a/src/app/api/waitlist/leaderboard/route.ts
+++ b/src/app/api/waitlist/leaderboard/route.ts
@@ -67,7 +67,7 @@ type LeaderboardResponse = {
 }
 
 const CACHE_KEY_NAMESPACE = 'waitlist:leaderboard'
-const CACHE_TTL_MS = Number(process.env.WAITLIST_LEADERBOARD_CACHE_MS ?? 15_000) // 15s default
+const CACHE_TTL_MS = Number(process.env.WAITLIST_LEADERBOARD_CACHE_MS ?? 120_000) // 120s default (increased from 15s)
 const CACHE_TTL_SECONDS = Math.max(1, Math.floor(CACHE_TTL_MS / 1000))
 const STALE_SECONDS = CACHE_TTL_SECONDS * 3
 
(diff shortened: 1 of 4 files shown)


SYMBaiEX (commented):
Looks good to me, great catch

PR #266, opened 2025-11-24
Fix points calculation to include base points

wtfsayo:
Updated the points calculation logic in ComingSoon to add base points to the total. This ensures consistency between the calculated and reported points values.


SYMBaiEX (commented):
This is good to go, this adds calculation to point mismatches as the base 1000 signup was not being calculated with the total points correctly.

PR #554, opened 2025-12-02
fix: Agent chat failing due to missing plugin registration

tcm390:
### Question

In this PR I removed the manual `modelDelegates` assignment and switched to using `runtime.registerPlugin()` instead.

**Previously we were doing:**
runtime.modelDelegates[type] = handler;

**Now we're doing:**
await runtime.registerPlugin(groqPlugin);

Just want to confirm was `modelDelegates` being used anywhere else or for any other purpose I might have missed? Want to make sure this change doesn't break anything downstream.

result:

Screenshot 2025-12-03 at 3 42 44 AM

### Problem

Agent chat was returning 500 errors with two distinct issues:

1. **"No handler found for delegate type: TEXT_LARGE"** - Model handlers weren't registered because we were manually setting `modelDelegates` instead of properly calling `runtime.registerPlugin()`.

2. **"Cannot read properties of undefined (reading 'log')"** - After fixing plugin registration, ElizaOS runtime's internal `useModel` method was calling `adapter.log` which was undefined.

### Solution

- **Removed manual `modelDelegates` assignment** - The runtime expects plugins to be registered via `registerPlugin()`.

- **Register all plugins via `runtime.registerPlugin()`** - Now properly registering `groqPlugin`, `experiencePlugin`, and `trajectoryLoggerPlugin` using the official API.

- **Override `adapter.log` with no-op** (expects `{ body, entityId, roomId, type }`).

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -29,6 +29,8 @@
     "@babylon/training": "workspace:*",
     "@elizaos/core": "^1.6.4",
     "@elizaos/plugin-sql": "^1.6.4",
+    "@elizaos/plugin-openai": "1.5.18",
+    "@elizaos/plugin-anthropic": "1.5.12",
     "@fal-ai/client": "^1.7.2",
     "@farcaster/auth-client": "^0.7.0",
     "@farcaster/auth-kit": "^0.8.1",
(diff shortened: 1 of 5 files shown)


lalalune:
Probably fine I think we want an env-dependent registration of plugin to use so the game can run on openai or claude as well if those keys are available

tcm390:
> Probably fine I think we want an env-dependent registration of plugin to use so the game can run on openai or claude as well if those keys are available

Got it! And we also have this optional provider param in useModel to select a specific LLM provider if we want to register multiple 

(https://github.com/elizaOS/eliza/blob/4e6fb8e8b3f2887bd775c8197e98d1fc1e2af9b5/packages/core/src/runtime.ts#L2118)

PR #664, opened 2025-12-23
feat(agents): Allow users to pick their own agent name and username

tcm390:
## Problem

https://github.com/user-attachments/assets/99d38e0b-cfa3-47c7-ab63-786b5dd5a5ff

User-created agents were not accessible via their profile pages. When clicking on an agent post in the feed, users would be redirected to a URL like `/profile/agent_tcm_posting_agent7_69zrch` and encounter the error:

> User or Actor "agent_tcm_posting_agent7_69zrch" not found

### Root Cause

The `UserIdSchema` validation in `packages/shared/src/validation/schemas/common.ts` had a 30-character limit for usernames. However, agent usernames were auto-generated using the format:

```
agent_${baseUsername}_${randomSuffix}
```

This resulted in usernames up to 33 characters:
- `agent_` prefix: 6 characters
- `baseUsername`: up to 20 characters
- `_` separator: 1 character
- `randomSuffix`: 6 characters

The profile page API calls would fail validation, causing agents to appear as "not found."

## Solution

Rather than simply increasing the character limit, we redesigned the agent creation UX based on team discussion:

1. **Let users pick their own agent name and username** at creation time
2. **Validate username uniqueness** in real-time (reusing existing `/api/onboarding/check-username` endpoint)
3. **Replace auto-generated usernames** with user-chosen ones that follow the same validation rules as regular users

### New Components

- **`AgentNameModal.tsx`**: Modal that prompts users to enter their agent's display name and username before proceeding to the full agent configuration page
- **`useAgentUsernameCheck.ts`**: Hook for real-time username availability checking with debouncing

### Before
1. User clicks "Create Agent"
2. System auto-generates name like "MuPrime-2856" and username like "agent_muprime_2856_x7f3k2"
3. Agent profile pages were inaccessible due to validation issues

### After
1. User clicks "Create Agent"
2. Modal appears prompting for display name and username
3. Real-time validation shows username availability
4. User proceeds to configure agent with their chosen identity
5. System prompt automatically uses the user's chosen name
6. Agent profile pages work correctly

## Screenshots

https://github.com/user-attachments/assets/8ae8b0c3-835d-4051-a24d-c91579ea4dd0

## Backward Compatibility

- Programmatic agent creation (e.g., training scripts) continues to work with auto-generated usernames when `username` is not provided
- Existing agents with legacy usernames are unaffected

## Summary by CodeRabbit

* **New Features**
  * Agents can have optional usernames shown as @username with a "View Profile" external link.
  * Creation flow now requires display name and username up front; username is submitted and persisted and validated server-side.
  * Live username availability checks with suggestions, retry, inline status, and minimum-length validation.

* **Refactor**
  * Profile setup modal renamed/repurposed and now appears first; profile preview is view-only and inline edit removed.


SYMBaiEX (approved):
Looks good, discussed over DM's.  Great implementation.

PR #669, opened 2025-12-24
refactor: unify group chat schema

tcm390:
## Summary

This PR consolidates the two separate group chat systems (user-created groups and NPC-managed groups) into a single unified schema. Previously, we had fragmented tables and inconsistent flows that were difficult to maintain and scale.

### The Problem: Two Parallel Systems

We had **two completely separate group systems** that evolved independently:

1. **User-Created Groups** (`UserGroup`, `UserGroupMember`, `UserGroupAdmin`, `UserGroupInvite`)
   - Created via the Groups API
   - Had invite flow for adding members
   - Stored admin roles in a separate table

2. **NPC-Managed Groups** (`GroupChatMembership`, `Chat.npcAdminId`)
   - No group table
   - Created by NPCs during game ticks
   - No invite flow - users were added directly
   - Quality tracking (qualityScore, messageCount) for kick mechanics
   - Used `Chat.npcAdminId` instead of a proper Group reference

### Issues with the Old System

- **Two admin systems**: `UserGroupAdmin` table + `ChatAdmin` (dead code, never used)
- **Inconsistent membership tracking**: `UserGroupMember` vs `GroupChatMembership` with different fields
- **Heuristic-based group type detection**: Admin panel used `chat.name.includes(npc.name)` to guess if a group was NPC-managed
- **User-created groups had validation issues**: Members weren't properly validated, leading to bugs (temporary fix in PR #665: https://github.com/BabylonSocial/babylon/pull/665)
- **Duplicate logic**: When a user left a chat, code had to update both `ChatParticipant` AND the appropriate membership table
- **Inconsistent invite behavior**: Adding members during group creation directly added them without invite/consent, while adding members after creation required invite acceptance

### Decision to Refactor

Rather than continuing to patch the fragmented system, I decided to consolidate everything into a unified schema that:
- Supports all group types (`user`, `npc`, `agent`) with the same tables
- Has a single invite flow for consistency
- Uses explicit `group.type` instead of heuristics
- Scales better and is easier to maintain

### New Unified Schema

```
Group
├── id, name, description
├── type: 'user' | 'npc' | 'agent'
├── ownerId (who controls it)
├── createdById, createdAt, updatedAt

GroupMember
├── groupId, userId
├── role: 'owner' | 'admin' | 'member'
├── qualityScore, messageCount, lastMessageAt (for NPC kick mechanics)
├── isActive, joinedAt, addedBy
├── kickedAt, kickReason

GroupInvite
├── groupId, invitedUserId, invitedBy
├── status: 'pending' | 'accepted' | 'declined'
├── message, invitedAt, respondedAt

Chat.groupId → Group.id (replaces npcAdminId)
```

## Testing

- Tested NPC group formation by temporarily setting `FORM_NEW_GROUP_CHANCE` to 100%
- Verified NPCs could invite users and users could accept/decline invites
- Verified NPC message posting works correctly with the new schema
- Tested group limit enforcement (max 5 NPC groups per user)

## Screenshot

https://github.com/user-attachments/assets/629b7c5c-5eed-4423-a4da-087f0e39f6ee

https://github.com/user-attachments/assets/53aa8780-c763-4767-9f09-d5cbde5cd31c

## Migration Notes

- **No data migration script included** - schema changes will be applied manually
- Old tables are removed from schema but may still exist in database until migration runs
- `Chat.npcAdminId` column removed - groups now use `Chat.groupId` → `Group.id`

## Known Issues

There are some UI issues related to the refactored group system that will be addressed in a follow-up PR.

## Removed Unused Code

- `apps/web/src/app/api/user-groups/` - entire directory (8 route files)
- `apps/web/src/components/groups/` - `GroupDetailsModal`, `GroupInviteNotification`, `GroupInvitesPanel`, `GroupSettingsModal`, `UserGroupsList`

## Summary by CodeRabbit

* **New Features**
  * Unified group system with owner/admin/member roles, tighter group–chat linking, invite flow updates, group-scoped notifications, and NPC membership limits.

* **Bug Fixes**
  * More reliable membership checks and stronger owner/admin protections for join/leave/promote/demote flows.

* **Refactor**
  * Consolidated invites and membership into a single group-based model; invite/accept flows now return richer info.

* **Breaking Changes**
  * Legacy user-group APIs and several group UI components removed or replaced.


packages/db/src/schema/messaging.ts
```diff
@@ -32,12 +35,11 @@ export const chats = pgTable(
     index('Chat_groupId_idx').on(table.groupId),
```
slkzgm, line 35: The comment says "one Chat per Group", but schema only has an index on `Chat.groupId`. Should we add a UNIQUE constraint on `Chat.groupId` (NULLs allowed) to enforce the invariant and avoid `findFirst({ where: { groupId } })` becoming non-deterministic?

packages/db/src/schema/messaging.ts
```diff
@@ -225,92 +144,112 @@ export const notifications = pgTable(
...
   {
     id: text('id').primaryKey(),
     groupId: text('groupId').notNull(),
     userId: text('userId').notNull(),
-    grantedAt: timestamp('grantedAt', { mode: 'date' }).notNull().defaultNow(),
-    grantedBy: text('grantedBy').notNull(),
+    // 'owner' = creator/controller of the group
+    // 'admin' = can manage members
+    // 'member' = regular participant
+    role: text('role').notNull().default('member'),
```
slkzgm, line 196: `GroupMember.role` is a free-form text column. Would we want a `pgEnum` (or CHECK constraint) here to prevent unexpected values from slipping in (esp. since access control depends on it)?

packages/db/src/schema/messaging.ts
```diff
@@ -225,92 +144,112 @@ export const notifications = pgTable(
...
+ */
+export const groupInvites = pgTable(
+  'GroupInvite',
   {
     id: text('id').primaryKey(),
     groupId: text('groupId').notNull(),
     invitedUserId: text('invitedUserId').notNull(),
     invitedBy: text('invitedBy').notNull(),
+    // 'pending' | 'accepted' | 'declined'
     status: text('status').notNull().default('pending'),
```
slkzgm, line 231: Same question for `GroupInvite.status`: should this be a `pgEnum` / CHECK constraint? Also, do we intend to allow re-invites by updating the existing row (single invite per group/user) or by allowing multiple historical invites (partial-unique on pending)?

packages/engine/src/services/alpha-group-invite-service.ts
```diff
@@ -116,15 +115,16 @@ export class AlphaGroupInviteService {
         continue;
       }
 
-      // Check if already invited to a group with this NPC
+      // Check if already in a group managed by this NPC
+      // The GroupChatService.recordInvite will handle duplicates,
+      // but we do a quick check here for efficiency
       const [existingMembership] = await db
```
slkzgm, line 120: This membership check looks broader than the comment "managed by this NPC": it currently skips users who are in *any* active group (including user groups). Is the intent instead to check membership in NPC groups owned by `npcId` (join `groups` + filter `groups.ownerId = npcId` / `groups.type = 'npc'`)?

packages/engine/src/services/alpha-group-invite-service.ts
```diff
@@ -162,27 +164,27 @@ export class AlphaGroupInviteService {
       // Check if user is in invite cooldown
```
slkzgm: Invite cooldown uses the latest membership across all groups. If the cooldown is meant for NPC invites only, should this mirror `NPCGroupDynamicsService.filterUsersForInvite` (join `groups` and filter `groups.type = 'npc'`)?

packages/engine/src/services/group-chat-service.ts
```diff
@@ -275,14 +285,45 @@ export class GroupChatService {
...
+        updatedAt: new Date(),
+      });
+
+      // Create chat with groupId link
       await db.insert(chats).values({
         id: chatId,
         name: chatName,
         isGroup: true,
         gameId: 'realtime',
+        groupId, // Link Chat → Group
```
slkzgm: Question on semantics: `recordInvite` still adds `ChatParticipant` + `GroupMember` immediately (no `GroupInvite`). After this refactor, are NPC invites supposed to require user acceptance? If yes, I’d expect this to create a `GroupInvite` and let the accept endpoint handle membership/participants.

packages/engine/src/services/group-chat-service.ts
```diff
@@ -275,14 +285,45 @@ export class GroupChatService {
...
+        updatedAt: new Date(),
+      });
+
+      // Update chat with groupId
+      await db.update(chats).set({ groupId }).where(eq(chats.id, chatId));
+    } else {
+      groupId = existingChat.groupId;
     }
 
     // Check if participant exists
```
slkzgm: Edge case: if a participant exists but `isActive = false`, we currently do nothing (and unique constraint prevents re-insert). Should we reactivate the participant here? Similar re-join handling likely needed for an existing inactive `GroupMember` (unique on groupId+userId).

apps/web/src/app/api/admin/group-invite/route.ts
```diff
@@ -195,12 +240,13 @@ export const POST = withErrorHandling(async (request: NextRequest) => {
     }
 
     // Record membership
```
slkzgm: `db.groupMember.create` can throw on the `(groupId,userId)` unique constraint if the user was previously a member (inactive) or if two invites race. Consider upserting/reactivating an existing `GroupMember` (and `ChatParticipant.isActive`) rather than unconditional create.

Also: `notifyGroupChatInvite` uses the global `db` (not the `asSystem` tx client). Calling it inside the transaction feels risky; would it be safer to run notifications after the transaction commits?

apps/web/src/app/api/groups/invites/[inviteId]/accept/route.ts
```diff
@@ -84,58 +46,119 @@ export const POST = withErrorHandling(
         throw new ApiError('This invite has already been processed', 400);
       }
 
-      // Check if user is already a member
-      const existingMember = await db.userGroupMember.findFirst({
+      // Check if user is at the NPC group limit (only NPC groups count toward the limit)
```
Comment (slkzgm) [inline apps/web/src/app/api/groups/invites/[inviteId]/accept/route.ts:54]: This enforces the NPC group cap without checking the invited group’s type. If a user is at the NPC cap but accepts a *user* group invite, should that be allowed? Might need to fetch the invited group and only apply the cap when `group.type === 'npc'`.

apps/web/src/app/api/groups/invites/[inviteId]/accept/route.ts
```diff
@@ -84,58 +46,119 @@ export const POST = withErrorHandling(
...
         where: {
           groupId: invite.groupId,
-          isGroup: true,
+          userId: user.userId,
         },
       });
 
-      if (chat) {
-        await db.chatParticipant.create({
+      if (existingMember) {
```
Comment (slkzgm) [inline apps/web/src/app/api/groups/invites/[inviteId]/accept/route.ts]: When `existingMember.isActive`, code updates the invite then throws. Because this runs inside the `asUser` transaction, that update will be rolled back, leaving the invite pending and effectively impossible to clear. Suggestion: make accept idempotent (mark accepted + ensure participant active + return 200) instead of throwing.

apps/web/src/app/api/groups/invites/[inviteId]/accept/route.ts
```diff
@@ -84,58 +46,119 @@ export const POST = withErrorHandling(
...
+          });
+        }
       }
 
       // Update invite status
-      await db.userGroupInvite.update({
+      await db.groupInvite.update({
         where: { id: inviteId },
         data: {
           status: 'accepted',
```
Comment (slkzgm) [inline apps/web/src/app/api/groups/invites/[inviteId]/accept/route.ts:162]: This marks *all* `group_invite` notifications as read for the user. Should this be scoped to `inviteId` (and/or `groupId`) to avoid clearing unrelated invites?

apps/web/src/app/api/groups/invites/[inviteId]/decline/route.ts
```diff
@@ -83,7 +46,7 @@ export const POST = withErrorHandling(
       }
 
       // Update invite status
-      await db.userGroupInvite.update({
+      await db.groupInvite.update({
         where: { id: inviteId },
         data: {
           status: 'declined',
```
Comment (slkzgm) [inline apps/web/src/app/api/groups/invites/[inviteId]/decline/route.ts:52]: Same concern as accept: `updateMany` marks all `group_invite` notifications read. Should this filter by the specific `inviteId`?

packages/mcp/src/handlers/tool-handlers.ts
```diff
@@ -1390,6 +1415,31 @@ export async function executeCreateGroup(
...
+  await db.groupMember.createMany({
+    data: [
+      {
+        id: memberIds[0]!,
+        groupId,
+        userId: agent.userId,
+        role: 'owner',
+        addedBy: agent.userId,
+      },
+      ...args.memberIds.map((memberId, idx) => ({
```
slkzgm: `args.memberIds` isn’t normalized (can include duplicates and/or `agent.userId`). That will violate unique constraints on `ChatParticipant(chatId,userId)` and `GroupMember(groupId,userId)` for `createMany`. Should we dedupe + filter before inserting?

packages/mcp/src/handlers/tool-handlers.ts
```diff
@@ -1475,65 +1548,139 @@ export async function executeMarkNotificationsRead(
...
-      groupId: invite.chatId,
-      groupName: chatsMap.get(invite.chatId)?.name || null,
-      inviterId: invite.invitedBy,
-      timestamp: invite.invitedAt.toISOString(),
-    })),
+    invites: invitesList.map((invite) => {
+      const group = groupsMap.get(invite.groupId);
+      return {
+        id: invite.id,
+        groupId: chatsMap.get(invite.groupId) || invite.groupId, // Return chatId for backward compat
```
slkzgm, line 1606: Returning `chatId` in a field named `groupId` ("for backward compat") seems risky for downstream consumers. Could we return both `groupId` and `chatId` (and keep `groupId` truthful)?

packages/mcp/src/handlers/tool-handlers.ts
```diff
@@ -1475,65 +1548,139 @@ export async function executeMarkNotificationsRead(
...
 /**
  * Execute accept_group_invite tool
+ * Chat.groupId → Group.id relationship
  */
 export async function executeAcceptGroupInvite(
   agent: AuthenticatedAgent,
   args: AcceptGroupInviteArgs
 ): Promise<AcceptGroupInviteResult> {
-  const invite = await db.chatInvite.findUnique({
+  const invite = await db.groupInvite.findUnique({
```
slkzgm, line 1625: Accept flow isn’t idempotent and isn’t transactional: no `invite.status === 'pending'` check, and it `create`s ChatParticipant/GroupMember unconditionally (unique constraints). If any insert fails after the invite update, state can be inconsistent. Suggest wrapping in a transaction and reusing the web route’s reactivation logic.

slkzgm (changes requested):
Thanks for taking on a big, high-impact refactor here — unifying the group schema is the right direction.

I reviewed the diff against `staging` and also skimmed the existing bot feedback. I think we’re close, but I’m requesting changes because there are a few correctness / consistency issues that look merge-blocking.

**Main blockers (please address before merge)**
1) **AlphaGroupInviteService eligibility checks** look too broad (users in *any* group can get excluded / cooled down for NPC invites).
2) **Invite semantics are inconsistent across the codebase** (some “invites” create membership immediately; others create `GroupInvite` and require acceptance). Please confirm the intended UX for NPC invites and align the implementations.
3) **Idempotency / uniqueness**: several flows can hit `GroupMember(groupId,userId)` and `ChatParticipant(chatId,userId)` unique constraints (notably MCP accept/create and admin NPC invite).
4) **Notifications**: accept/decline currently mark *all* `group_invite` notifications as read (likely too broad).

**Non-blocking / follow-ups**
- Consider DB-level constraints for `GroupMember.role` and `GroupInvite.status` (enum / CHECK), and enforcing the “one chat per group” invariant.
- Migration plan: PR description says there’s no migration script. I’d strongly prefer we ship a Drizzle migration (or at least a documented, repeatable migration procedure) so deploys don’t rely on manual steps.

Verdict: **Request changes** (see inline comments for specifics).

PR #670, opened 2025-12-24
feat(engine): Add tiered group system for scalable NPC group access

puncar-dev:
## Summary

- **Implement 3-tier NPC group system**: Inner Circle (12 max), Community (50 max), Followers (500 max) per NPC
- **Scale capacity**: From 1,680 total users to 78,680 (46x increase for 140 NPCs)
- **Asymmetric information mechanics**: Tier-specific alpha sharing (full → partial → public)

## Type

- [x] Feature

## Context / Links

- Addresses scalability for 300,000 users
- Builds on existing NPC group dynamics

## Scope

- [x] Domain / game engine (`packages/engine`)
- [x] DB (`packages/db`)
- [x] Web API routes (`apps/web`)
- [x] Docs

### New Files
- `packages/engine/src/services/tier-config.ts` - TIER_CONFIG constants and helpers
- `packages/engine/src/services/tiered-group-service.ts` - Core service
- `packages/engine/src/prompts/game/group-messages-tier{1,2,3}.ts` - Tier-specific prompts
- `packages/db/drizzle/migrations/0011_add_tiered_group_system.sql` - Migration
- `packages/db/drizzle/migrations/0012_rollback_tiered_group_system.sql` - Rollback
- `apps/web/src/app/api/tiers/` - User tier status endpoints
- `apps/web/src/app/api/admin/tiers/` - Admin analytics endpoint
- `packages/testing/integration/tiered-group-system.test.ts` - 63 tests

### Modified Files
- `npc-group-dynamics-service.ts` - Integrated tiered invitations and promotions
- `notification-service.ts` - Added tier promotion/demotion notifications
- `ChatListItem.tsx` - Added tier badges (Gold/Blue/Gray)
- Chat API routes - Return tier info

## Review guide

**Start here**: `packages/engine/src/services/tier-config.ts` for tier definitions, then `tiered-group-service.ts` for core logic.

**Risk level**: Medium - DB migration required

## Test plan

- [x] `bun run lint` (our files pass, pre-existing errors in staging)
- [x] `bun test packages/testing/integration/tiered-group-system.test.ts` (63 pass)
- [ ] `bun run typecheck` (pre-existing errors in babylon-typescript-agent-example)

- **DB migration required**: `bun run db:migrate` to apply 0011
- **Rollback available**: 0012_rollback_tiered_group_system.sql
- **No env changes**

## Breaking changes

None - existing groups continue to work, will be migrated to Tier 1 by migration

## Security / privacy

No impact

## Screenshots

No UI changes beyond tier badges in chat list (Gold shield = Inner Circle, Blue = Community, Gray = Followers)

## Checklist (author)

- [x] Self-review done
- [x] Base branch correct (`staging`)
- [x] No screenshots needed (minor badge addition)
- [x] Handlers thin and portable
- [x] Domain logic in packages
- [x] Tests added (63 tests covering config, helpers, and service methods)
- [x] Code owners requested

## Summary by CodeRabbit

* **New Features**
  * Tiered group system with three engagement levels (Inner Circle, Community, Followers).
  * Automated invites, promotions, and demotions based on engagement and activity rules.
  * Tier-specific membership limits, messaging frequency, and guidance for richer NPC interactions.
  * New user-facing views: per-NPC tier listings, individual tier status, and global tier analytics/capacity insights.
  * Existing NPC groups and memberships initialized into Tier 1 (Inner Circle) to preserve continuity.

### DIFF
--- a/packages/engine/src/services/tiered-group-service.ts
+++ b/packages/engine/src/services/tiered-group-service.ts
@@ -19,10 +19,12 @@ import {
   eq,
   groupMembers,
   groups,
+  inArray,
   isNotNull,
   isNull,
   ne,
 } from '@babylon/db';
+import { DistributedLockService } from '@babylon/api';
 import { GROUP_CONFIG, generateSnowflakeId, logger } from '@babylon/shared';
 
 import { NPCInteractionTracker } from './npc-interaction-tracker';
@@ -35,6 +37,7 @@ import {
   getTierForEngagementScore,
   getTierGroupName,
   isEligibleForPromotion,
+  isValidTier,
   shouldDemote,
   TIER_CONFIG,
   type TierLevel,
@@ -95,7 +98,7 @@ export class TieredGroupService {
 
     const existingTierMap = new Map(
       existingGroups
-        .filter((g) => g.tier !== null)
+        .filter((g) => isValidTier(g.tier))
         .map((g) => [g.tier as TierLevel, g])
     );
 
@@ -252,12 +255,22 @@ export class TieredGroupService {
         .where(eq(chats.groupId, g.id))
         .limit(1);
 
-      const config = getTierConfig(g.tier as TierLevel);
+      // Validate tier before processing
+      if (!isValidTier(g.tier)) {
+        logger.warn(
+          `Invalid tier value ${g.tier} for group ${g.id}, skipping`,
+          { groupId: g.id, tier: g.tier },
+          'TieredGroupService'
+        );
+        continue;
+      }
+
+      const config = getTierConfig(g.tier);
       const maxMembers = g.maxMembers ?? config.maxMembers;
       const memberCount = countResult?.count ?? 0;
 
       result.push({
-        tier: g.tier as TierLevel,
+        tier: g.tier,
         groupId: g.id,
         chatId: chat?.id ?? null,
         groupName: g.name,
@@ -306,8 +319,8 @@ export class TieredGroupService {
     let canBePromoted = false;
     let promotionBlockedReason: string | null = null;
 
-    if (membership?.tier !== null && membership?.tier !== undefined) {
-      const currentTier = membership.tier as TierLevel;
+    if (isValidTier(membership?.tier)) {
+      const currentTier = membership.tier;
       const daysInTier = Math.floor(
         (Date.now() - membership.joinedAt.getTime()) / (1000 * 60 * 60 * 24)
       );
@@ -345,7 +358,7 @@ export class TieredGroupService {
     return {
       userId,
       npcId,
-      currentTier: (membership?.tier as TierLevel) ?? null,
+      currentTier: isValidTier(membership?.tier) ? membership.tier : null,
       groupId: membership?.groupId ?? null,
       joinedAt: membership?.joinedAt ?? null,
       engagementScore,
@@ -357,12 +370,30 @@ export class TieredGroupService {
 
   /**
    * Invite user to appropriate tier based on engagement
+   *
+   * Uses distributed locking to prevent race conditions where multiple
+   * concurrent invites could exceed group capacity.
    */
   static async inviteUserToTier(
     userId: string,
     npcId: string
   ): Promise<{ success: boolean; tier: TierLevel | null; reason: string }> {
-    // Check if already in a tier with this NPC
+    // Validate that npcId is a valid NPC
+    const actor = StaticDataRegistry.getActor(npcId);
+    if (!actor) {
+      logger.warn(
+        `inviteUserToTier called with invalid NPC ID: ${npcId}`,
+        { userId, npcId },
+        'TieredGroupService'
+      );
+      return {
+        success: false,
+        tier: null,
+        reason: `Invalid NPC ID: ${npcId}`,
+      };
+    }
+
+    // Check if already in a tier with this NPC (fast-fail before lock)
     const [existing] = await db
       .select({ id: groupMembers.id })
       .from(groupMembers)
@@ -385,7 +416,7 @@ export class TieredGroupService {
       };
     }
 
-    // Check group limit
+    // Check group limit (fast-fail before lock)
     const [groupCount] = await db
       .select({ count: count() })
       .from(groupMembers)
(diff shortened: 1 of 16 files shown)


packages/engine/src/services/tiered-group-service.ts
```diff
@@ -0,0 +1,1085 @@
...
+            promotionBlockedReason = `Need engagement score ${needed}+ (current: ${engagementScore.toFixed(0)})`;
+          }
+        }
+      }
+    }
+
+    return {
+      userId,
+      npcId,
+      currentTier: membership?.tier as TierLevel | null,
```
slkzgm: Potential runtime shape issue: `currentTier: membership?.tier as TierLevel | null` can become `undefined` when `membership` is missing. Would you switch to something like `currentTier: (membership?.tier ?? null) as TierLevel | null` (and maybe validate 1/2/3) so the API always returns `null` when the user is not in any tier?

packages/engine/src/services/tiered-group-service.ts
```diff
@@ -0,0 +1,1085 @@
...
+      );
+
+    if ((groupCount?.count ?? 0) >= MAX_ACTIVE_USER_GROUPS) {
+      return {
+        canInvite: false,
+        reason: `At maximum of ${MAX_ACTIVE_USER_GROUPS} groups`,
+      };
+    }
+
+    // Check cooldown
```
slkzgm: Question on cooldown semantics: this check looks at any `GroupChatMembership` joined in the last `INVITE_COOLDOWN_HOURS` regardless of `isActive` (and regardless of NPC). In staging, the cooldown logic seemed based on latest active membership. Is the stricter global cooldown intentional? If not, consider filtering by `isActive = true` (and/or by `npcAdminId`).

packages/engine/src/services/tiered-group-service.ts
```diff
@@ -0,0 +1,1085 @@
...
+        success: false,
+        tier: null,
+        chatId: null,
+        chatName: null,
+        reason: `No available tier (score: ${engagementScore.toFixed(0)}, min required: ${TIER_CONFIG[3].minEngagementScore})`,
+      };
+    }
+
+    // Add to chat participants
+    await db.insert(chatParticipants).values({
```
slkzgm: Requesting clarification: `ChatParticipant` is unique on `(chatId, userId)` and `GroupChatMembership` is unique on `(userId, chatId)`. If a user ever gets removed from a tier and later re-enters the same tier (e.g., Tier 2 -> Tier 3 -> Tier 2), these inserts will likely conflict unless we reactivate existing rows. Would an `onConflictDoUpdate` (or an update-then-insert fallback) be safer here? Also worth wrapping both inserts in a transaction.

packages/engine/src/services/tiered-group-service.ts
```diff
@@ -0,0 +1,1085 @@
...
+    const higherTier = getHigherTier(status.currentTier);
+    if (!higherTier) return null;
+
+    const tiers = await this.getNpcTiers(npcId);
+    const targetTierInfo = tiers.find((t) => t.tier === higherTier);
+    if (!targetTierInfo || targetTierInfo.isFull) {
+      return null;
+    }
+
+    // Remove from current tier
```
slkzgm: Promotion path updates 2 tables then inserts 2 rows. Without a transaction, a mid-way failure can leave the user removed from the old tier but not added to the new one. Also, the inserts into `targetTierInfo.chatId` will conflict if the user was previously in that tier chat (unique constraints). Can we make this atomic and re-activate existing participant/membership rows instead of inserting unconditionally?

packages/engine/src/services/tiered-group-service.ts
```diff
@@ -0,0 +1,1085 @@
...
+  ): Promise<DemotionResult | null> {
+    const status = await this.getUserTierStatus(userId, npcId);
+
+    if (!status.currentTier || !status.chatId) {
+      return null;
+    }
+
+    const lowerTier = getLowerTier(status.currentTier);
+
+    // Remove from current tier
```
slkzgm: Similar concern on demotion: we deactivate the current tier before confirming the target tier exists/has space, and then we insert into the lower tier. If the lower-tier inserts fail (or the tier chat is missing), the user ends up in no tier. Would you consider (1) checking target tier availability first, (2) wrapping the whole move in a transaction, and (3) reactivating existing participant/membership rows for the lower tier to avoid unique constraint conflicts?

packages/engine/src/services/tiered-group-service.ts
```diff
@@ -0,0 +1,1085 @@
...
+    if (actors.length === 0) {
+      logger.debug(
+        'No actors in StaticDataRegistry, skipping tier invitations',
+        undefined,
+        'TieredGroupService'
+      );
+      return { invites, errors };
+    }
+
+    for (const actor of actors) {
```
slkzgm: This runs for every actor on every tick and calls `ensureAllTiersExist` + `getEligibleUsersForNpc` (which computes engagement scores). At 100+ NPCs this could get expensive fast. Is this intentionally per-tick, or should invites run at a lower frequency / for a subset of NPCs, or use cached/precomputed engagement rankings?

packages/engine/src/services/tiered-group-service.ts
```diff
@@ -0,0 +1,1085 @@
...
+        )
+      )
+      .where(
+        and(
+          eq(users.isActor, false),
+          isNotNull(users.walletAddress),
+          isNull(groupChatMemberships.id) // Not in any group with this NPC
+        )
+      )
+      .limit(50); // Reduce to 50 since we calculate engagement for each
```
slkzgm: `limit(50)` without an explicit order makes the candidate pool effectively arbitrary (and not truly "top engaged" users). If the intent is sampling, maybe add a deterministic/random order; if the intent is "top engaged", we probably need a cheaper prefilter or a precomputed score table/index so we do not compute engagement for a mostly random slice.

packages/db/drizzle/migrations/0011_add_tiered_group_system.sql
```diff
@@ -0,0 +1,38 @@
...
+SET 
+  "tier" = 1,
+  "tierName" = 'Inner Circle',
+  "maxMembers" = 12
+WHERE "isGroup" = true 
+  AND "npcAdminId" IS NOT NULL
+  AND "tier" IS NULL;
+
+-- Update existing memberships to Tier 1
+UPDATE "GroupChatMembership"
```
slkzgm: Migration question: `UPDATE "GroupChatMembership" SET "tier" = 1 WHERE "tier" IS NULL;` is very broad. Are we 100% sure `GroupChatMembership` only tracks NPC groups? If not, safer would be to join `Chat` and scope to chats that were migrated to tiered NPC groups (e.g., `Chat.isGroup = true AND Chat.npcAdminId IS NOT NULL`).

apps/web/src/app/api/chats/route.ts
```diff
@@ -425,6 +425,11 @@ export const GET = withErrorHandling(async (request: NextRequest) => {
           qualityScore: membership.qualityScore,
           lastMessageAt: membership.lastMessageAt,
           updatedAt: chat.updatedAt,
+          // Tier information
```
slkzgm: Nice adding tier metadata here. One consistency question: the `?all=true` public branch still returns chats without `tier/tierName/maxMembers/npcAdminId` (CodeRabbit flagged this too). If the discovery UI consumes these fields, should we include them there as well (even if `null`)?

apps/web/src/components/chats/ChatListItem.tsx
```diff
@@ -39,8 +71,31 @@ export function ChatListItem({
     >
       <div className="flex items-center gap-3">
         {chat.isGroup ? (
-          <div className="chat-button flex h-10 w-10 shrink-0 items-center justify-center rounded-full bg-sidebar-accent/50">
-            <Users className="h-5 w-5 text-primary" />
+          <div className="relative">
+            <div
+              className={cn(
+                'chat-button flex h-10 w-10 shrink-0 items-center justify-center rounded-full',
+                tierBadge ? 'bg-sidebar-accent/50' : 'bg-sidebar-accent/50'
```
slkzgm: Nit: `tierBadge ? 'bg-sidebar-accent/50' : 'bg-sidebar-accent/50'` is a no-op. If there is no intended visual difference, we can drop the conditional; if there is, we probably want different classes.

apps/web/src/components/chats/ChatListItem.tsx
```diff
@@ -39,8 +71,31 @@ export function ChatListItem({
...
+              )}
+            >
+              {tierBadge ? (
+                <tierBadge.Icon
+                  className={cn('h-5 w-5', tierBadge.className.split(' ')[1])}
```
slkzgm: Minor robustness: deriving icon color via `tierBadge.className.split(' ')[1]` is brittle (depends on class ordering). Could we store an explicit `iconClassName` (or use `text-*` separately) to avoid accidental breakage?

packages/api/src/services/notification-service.ts
```diff
@@ -638,3 +638,87 @@ export async function notifyGroupChatMessage(
...
+ */
+export async function notifyTierPromotion(
+  userId: string,
+  npcId: string,
+  chatId: string,
+  _chatName: string,
+  fromTier: number,
+  toTier: number
+): Promise<void> {
+  const result = await db
```
slkzgm: Suggestion: both functions query the NPC name on every call and repeat the tier-name mapping. If promotions/demotions happen in batch, this is N queries. Could we pass `npcName` from the caller (or prefetch once per NPC) and reuse `TIER_CONFIG`/a shared helper for tier display names to avoid drift?

apps/web/src/app/api/tiers/[npcId]/route.ts
```diff
@@ -0,0 +1,33 @@
...
+import { authenticate, successResponse, withErrorHandling } from '@babylon/api';
+import { TieredGroupService } from '@babylon/engine';
+import type { NextRequest } from 'next/server';
+
+export const GET = withErrorHandling(
+  async (
+    request: NextRequest,
+    context: { params: Promise<{ npcId: string }> }
+  ) => {
+    const { npcId } = await context.params;
```
Comment (slkzgm) [inline apps/web/src/app/api/tiers/[npcId]/route.ts]: Should we validate `npcId` upfront (e.g., ensure it is a known actor) and return 404/400? Right now we compute engagement and return empty tiers for unknown ids, which may be fine but I want to confirm expected behavior.

slkzgm (changes requested):
Thanks for the thorough PR + docs/tests — the overall structure (tier config + service layer + thin routes) is heading in a good direction.

## Requested changes (blockers)

1) **Tier transitions vs unique constraints**
I believe tier transitions will hit unique constraints in real promotion/demotion cycles:
- `ChatParticipant` is unique on `(chatId, userId)`
- `GroupChatMembership` is unique on `(userId, chatId)`

But `inviteUserToTier` / `promoteUser` / `demoteUser` mark the current rows inactive and then unconditionally `insert` into the target tier chat. A simple Tier 3 → Tier 2 → Tier 3 path would attempt to insert into the Tier 3 chat again while the old row still exists (inactive), which should fail.

2) **Atomicity for promotions/demotions**
Promotions/demotions are multi-table moves (update membership + update participants + insert participants + insert membership). Without a transaction, any mid-way failure can leave users in inconsistent states.

3) **`getUserTierStatus` response shape**
`currentTier: membership?.tier as TierLevel | null` can become `undefined` when the user has no membership; I think we should normalize to `null`.

## Major / follow-ups

- **Per-tick performance**: `processTickInvitations` runs `ensureAllTiersExist` + engagement score computation for each actor. I would like to confirm tick frequency and scale assumptions (this could become expensive at 100+ NPCs).
- **API shape consistency**: `GET /api/chats?all=true` still returns public chats without tier metadata (not in this diff hunk, but it may matter for chat discovery UI).

## Minor

- UI tier badge implementation has a couple small cleanups (redundant ternary, class split fragility).
- Tier promotion/demotion notifications duplicate tier name mapping and do an NPC lookup per call (could be optimized / de-duplicated).

**Verdict: Request changes** (mainly due to unique-constraint + atomicity concerns).

PR #674, opened 2025-12-28
fix: fail fast when NPC actorState missing instead of masking with default

wtfsayo:
## Summary
- When an NPC is missing their actorState record, now throws an error instead of silently using a default balance
- This enables fail-fast behavior for data issues while keeping the code simple

## Problem
The previous code masked data issues by silently using a default balance when an NPC had no actorState record. This made it harder to detect bootstrap problems.

## Solution
- If no actor record found: throw error with clear message to fix NPC bootstrap
- Simplified: removed redundant null check on `tradingBalance` since schema enforces `.notNull().default('10000')`
- Just use `Number(actor.tradingBalance)` directly

## Test plan
- [x] Run typecheck: `bun run typecheck`
- [x] Run lint: `bun run lint`
- [ ] Verify an NPC without actorState record causes MultiStepExecutor to throw with a clear error message
- [ ] Verify an NPC with valid actorState works as before

### DIFF
--- a/packages/agents/src/autonomous/MultiStepExecutor.ts
+++ b/packages/agents/src/autonomous/MultiStepExecutor.ts
@@ -277,14 +277,24 @@ export class MultiStepExecutor {
         .where(eq(actorState.id, agentUserId))
         .limit(1);
 
-      if (!actor?.tradingBalance) {
+      if (!actor) {
+        // Missing actorState record is a data issue that should be fixed at bootstrap
+        throw new Error(
+          `NPC ${agentUserId} has no actorState record. Run NPC bootstrap to create it.`
+        );
+      }
+
+      if (actor.tradingBalance === null || actor.tradingBalance === undefined) {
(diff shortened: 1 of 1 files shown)


wtfsayo:
@claude Thanks for the review! Regarding the null check concern:

The schema enforces `.notNull().default('10000')` on `tradingBalance` (packages/db/src/schema/actor-state.ts:23-25), so null values should never exist in the database.

Following the CLAUDE.md "fail fast" principle, we intentionally removed the defensive null check since:
1. Schema guarantees non-null values
2. If null somehow occurs, it indicates data corruption that should surface immediately rather than silently degrade to 0

The `Number(actor.tradingBalance)` approach aligns with the fail-fast philosophy - if the invariant is violated, we'll see the bug immediately rather than masking it.

Do you still recommend adding a defensive null check given the schema constraint, or is the current approach acceptable?

PR #675, opened 2025-12-28
fix: charge points after lock acquisition to prevent free actions on errors

wtfsayo:
## Summary
- Moved points deduction to immediately AFTER lock acquisition but BEFORE tick execution
- This ensures points are always charged once we commit to running an agent's tick
- Prevents abuse where agents could get free actions by causing errors during execution

## Rationale
Previously, points were deducted after `executeAutonomousTick()`. If that threw an exception, the catch block was hit and points were never deducted - agents got free actions on errors.

Now the flow is:
1. Acquire lock for agent
2. **Deduct points immediately** (commit to charging)
3. Get runtime
4. Execute autonomous tick
5. Process results
6. Release lock

## Test plan
- [x] Run typecheck: `bun run typecheck`
- [x] Run lint: `bun run lint`
- [ ] Verify points are deducted even if tick execution fails
- [ ] Verify lock is always released in finally block

### DIFF
--- a/apps/web/src/app/api/cron/agent-tick/route.ts
+++ b/apps/web/src/app/api/cron/agent-tick/route.ts
@@ -348,11 +348,21 @@ export async function POST(_req: NextRequest) {
       continue;
     }
 
+    // Always 1pt per tick for USER agents
+    const pointsCost = 1;
+
+    // CRITICAL: Deduct points immediately after lock acquisition, BEFORE tick execution.
+    // This ensures points are always charged once we commit to running the tick.
+    // If we deducted after tick execution, errors in executeAutonomousTick() would
+    // skip the deduction (catch block), allowing agents to get free actions on errors.
+    await agentService.deductPoints(
+      eligibleAgent.user.id,
+      pointsCost,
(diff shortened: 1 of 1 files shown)


wtfsayo:
@claude Thanks for the detailed review! The "inverted" logic is actually **intentional** based on a design decision:

**We chose Option B: Always charge after lock acquisition**

Rationale:
1. **Prevents abuse**: Users can't spam actions hoping for errors to get free ticks
2. **Simple accounting**: Every lock = 1 point (predictable)
3. **Rate limiting**: Acts as a natural rate limiter even on failures
4. **Vercel billing analogy**: Just like Vercel charges for function execution time regardless of errors, we charge for the tick slot

The original PR description was misleading (said "after execution") - I've updated it to correctly say "after lock acquisition, before execution".

The current implementation:
```
1. Acquire lock ✓
2. Deduct points ✓ (committed to this tick)
3. Execute tick (may fail)
4. Release lock
```

Is this approach acceptable given the intentional design choice, or do you see issues with charging upfront?

PR #713, opened 2026-01-02
BAB-96 fix agents treat amount as shares

wtfsayo:
## Type
- [x] Bug fix
- [ ] Feature
- [ ] Chore

## Context / Links
- [BAB-96](https://linear.app/eliza-labs/issue/BAB-96/fix-agents-trading-whole-shares-instead-of-dollar-amounts)
- Agents were outputting `amount` as share counts. We now require `amount_in_points`, provide balance context, and validate executor inputs.

## Scope
- `packages/agents/src/autonomous/AutonomousTradingService.ts`
- `packages/agents/src/autonomous/DirectExecutors.ts`

## Changes
1. Add balance-aware suggested trade size (10% of balance with $10 floor) and prompt copy so LLMs size positions in Babylon Points.
2. Normalize/validate `amount_in_points` vs the legacy `amount` field before executing trades.
3. Enhance executor warnings so “share-like” values only trigger when they are tiny relative to the user’s balance, reducing false alarms.

## Review Guide
- Start in `AutonomousTradingService` to review the prompt changes and parsing guard rails.
- Then check `DirectExecutors` for the refined heuristics that log suspicious share-sized trades.

## Test Plan
- [x] `turbo typecheck`
- [x] `turbo lint`
- [x] `turbo build`
- [x] `bunx biome lint packages/agents/src/autonomous/AutonomousTradingService.ts packages/agents/src/autonomous/DirectExecutors.ts`

## Ops / Migration
- [x] Not required

## Breaking Changes
- [x] None

## Security
- [x] No new surface area

## Screenshots / Recordings
- Backend-only change; no UI to capture.

## Summary by CodeRabbit

* **New Features**
  * Added recommended trade size calculation (10% of balance, bounded by $10 minimum and total balance) for autonomous trading.
  * Trade amounts now processed using Babylon Points as the primary indicator.

* **Improvements**
  * Enhanced trade amount validation with detection and warning for anomalous input patterns.

### DIFF
--- a/packages/agents/src/autonomous/AutonomousTradingService.ts
+++ b/packages/agents/src/autonomous/AutonomousTradingService.ts
@@ -147,6 +147,12 @@ export class AutonomousTradingService {
       })
       .join('\n');
 
+    const suggestedTradeSize =
+      balance.balance > 0
+        ? Math.min(Math.max(balance.balance * 0.1, 5), balance.balance)
+        : 0;
+    const suggestedTradeSizeText = suggestedTradeSize.toFixed(2);
+
     // Build trading prompt
     const prompt = `${config?.systemPrompt ?? 'You are an AI trading agent on Babylon.'}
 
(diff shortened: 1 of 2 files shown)


wtfsayo:
Addressed review feedback:
1. Suggested trade size now uses a  floor (matching backend min order) and the prompt copy calls that out.
2. The share-sized warning in `executeDirectTrade` now factors in account balance so it only fires when the trade is both 1-10 pts and <1% of balance, reducing false positives.

Updated the PR description to satisfy the template checklist as well.

SYMBaiEX (approved):
The core fix is solid, the defense-in-depth approach is appropriate, and you already addressed the key review feedback. The missing test coverage should be added as a follow-up.

PR #784, opened 2026-01-14
feat: add API key revocation for external agents

wtfsayo:
## Summary
- Add ability to revoke external agent API keys
- Track who registered each agent for ownership validation
- Skip revoked agents during API key verification

## Changes
- `packages/db/src/schema/agents.ts`: Added `revokedAt`, `revokedBy`, `registeredByUserId` fields + index on `revokedAt`
- `packages/agents/src/services/agent-registry.service.ts`: Added `revokeExternalAgent()` method, skip revoked in verification
- `apps/web/src/app/api/agents/external/[externalId]/revoke/route.ts`: New DELETE endpoint
- `apps/web/src/app/api/agents/external/register/route.ts`: Now tracks `registeredByUserId`

## Fixes Applied (from code review)
- ✅ **Legacy agents**: Added explicit null check - legacy agents (registered before this feature) can only be revoked by admins since ownership cannot be verified
- ✅ **Removed redundant check**: Route no longer duplicates `revokedAt` check (handled by service layer to avoid race conditions)
- ✅ **Added DB index**: Index on `revokedAt` for query performance

## Migration Required
Run `bun run db:generate` after merge to create migration for new schema fields.

## Test plan
- [ ] Register external agent, note API key works
- [ ] Call DELETE `/api/agents/external/{id}/revoke`
- [ ] Verify API key no longer works (401)
- [ ] Verify non-owner cannot revoke (403)
- [ ] Verify already-revoked returns appropriate error
- [ ] Verify legacy agent (null registeredByUserId) requires admin to revoke

## Summary by CodeRabbit

* **New Features**
  * External agent revocation: owners and administrators can now revoke external agent connections.
  * Agent ownership tracking: system records which user registered each external agent.

* **Security**
  * Enhanced API key verification: revoked external agent keys are now blocked from accessing the system.

### DIFF
--- a/apps/web/src/app/api/agents/external/[externalId]/revoke/route.ts
+++ b/apps/web/src/app/api/agents/external/[externalId]/revoke/route.ts
@@ -45,7 +45,11 @@ export async function DELETE(req: NextRequest, { params }: RouteParams) {
 
     // Check authorization: only owner or admin can revoke
     // Note: revokedAt check is handled by the service layer to avoid race conditions
-    const isOwner = connection.registeredByUserId === authUser.userId;
+    // Legacy agents (registered before this feature) have null registeredByUserId,
+    // meaning only admins can revoke them since ownership cannot be verified
+    const isOwner =
+      connection.registeredByUserId !== null &&
+      connection.registeredByUserId === authUser.userId;
     const isAdmin = await isUserAdmin(authUser.userId);
 
     if (!isOwner && !isAdmin) {
--- a/apps/web/src/app/api/agents/external/[externalId]/revoke/route.ts
+++ b/apps/web/src/app/api/agents/external/[externalId]/revoke/route.ts
@@ -9,7 +9,7 @@
  */
 
 import { agentRegistry } from '@babylon/agents';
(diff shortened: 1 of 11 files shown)


slkzgm (approved):
Overall this is a solid, security-relevant addition and it matches the PR description (revocation + ownership tracking + skipping revoked keys).

I pushed a small follow-up commit (`f8befae7f`) to tighten a few edge-cases:
- Added missing DB migrations for `registeredByUserId`, `revokedAt`, `revokedBy`.
- Made `revokeExternalAgent()` race-safe (atomic update) and ensured the route maps "already revoked" to `400` instead of leaking as `500` on concurrent requests.
- Ensured `/api/agents/external/connect` also rejects revoked keys and validates keys via the registry (previously it didn’t handle encrypted credentials).
- Filtered revoked connections at query level in `verifyExternalAgentApiKey()` and made credential parsing resilient.

Follow-ups (nice-to-have):
- Add automated tests around revoke/verify (owner vs admin vs non-owner; revoked keys return 401).
- If external agent count grows, consider optimizing API-key verification to avoid scanning/decrypting all connections per request.

PR #930, opened 2026-02-04
fix: use client-side Privy sendTransaction for buy points

revlentless:


slkzgm (dismissed):
## Summary
Switches the buy-points + admin ETH send flows to Privy’s client-side `useSendTransaction` with gas sponsorship (`sponsor: true`), avoiding the server-side `authorization_context.user_jwts` approach that isn’t working reliably with embedded wallets.

## What looks good
- Uses Privy’s client SDK for embedded wallets (better reliability, less server coupling).
- Keeps `ensureFunds` in place (sponsorship covers gas only, not the ETH value transfer).
- BigInt `value` usage stays type-safe.

## Notes / follow-ups
- The PR description is currently empty — please paste the rationale from the commit message so the context is captured on the PR.
- As noted in the commit message, other server-side sponsored flows (e.g. minting/betting) still use the old pattern; recommend tracking a follow-up to migrate them.

## Small fix pushed
I pushed commit `50ab56749` to reduce JWT diagnostics logging (debug level + no `sub`/`sid`) in `sendSponsoredEvmTransaction`.

## Verification
- `bun run typecheck` ✅
- `bun run lint` ✅
- `bun run build` ✅ (requires `NEXT_PUBLIC_CHAIN_ID` and an LLM key like `GROQ_API_KEY` in the environment)
- `bun run test:unit` currently fails in this branch due to missing exports from `packages/db` (appears unrelated to this PR).

PR #935, opened 2026-02-04
feat: improve profile replies and trades tab styling

jasperdisney:
## Summary
- **Trades tab**: Remove avatar and name row from trade cards, move timestamp inline with the trade action row for a cleaner layout
- **Replies tab**: Align ProfileReplyCard styling with feed's CommentPreview (avatar size md, header layout with justify-between, text-[15px] sizing, use shared CommentInteractionBar)
- **Replies tab**: Add interaction bar to the parent post/comment that was replied to, with full repost support for posts and like/reply for comments
- **API**: Fetch interaction counts (likes, comments, shares, isLiked, isShared) for parent posts and parent comments in the replies endpoint

### DIFF
--- a/apps/web/src/app/api/users/[userId]/posts/route.ts
+++ b/apps/web/src/app/api/users/[userId]/posts/route.ts
@@ -267,6 +267,149 @@ export const GET = withErrorHandling(
         );
       }
 
+      // Fetch interaction counts for parent posts
+      const [postLikeCounts, postCommentCounts, postShareCounts] =
+        await Promise.all([
+          db
+            .select({
+              postId: reactions.postId,
+              count: count(),
+            })
+            .from(reactions)
+            .where(
+              and(
+                inArray(reactions.postId, postIds),
+                eq(reactions.type, 'like')
+              )
+            )
+            .groupBy(reactions.postId),
+          db
+            .select({
+              postId: comments.postId,
+              count: count(),
+            })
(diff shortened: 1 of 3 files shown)


slkzgm (approved):
Overall looks solid and matches the PR description (UI parity with feed + parent interaction bars + counts on replies). Nice job keeping it batch-query based (no N+1).

I pushed a small follow-up commit to the PR branch to tighten perf/correctness:
- `9ff051ec9`: memoizes `InteractionBar` props in `ProfileReplyCard` to avoid re-running `InteractionBar`'s `useEffect` on every render (object identity churn).
- Excludes `deletedAt` comments from comment/reply counts in `/api/users/[userId]/posts?type=replies` (and the posts `commentCount`), so counts match other endpoints that ignore deleted comments.
- Reuses the formatted timestamp once per `TradeCard`.

Notes / concerns:
- API: adding parent post/comment counts adds several extra queries. They're all batched and `Promise.all`'d so it should be fine for `limit(100)`, but if this grows consider consolidating counts into fewer grouped queries.
- Local build: `bun run build` fails in this environment due to missing runtime config (LLM key + network/contract deployment constraints). Typecheck + lint pass.
- Unit tests are currently failing on `staging` as well (missing exports from `@babylon/db`), unrelated to this PR.

Given the above, LGTM after the follow-up commit.

PR #998, opened 2026-02-07
fix: prefer fresh Privy JWT over stale cookie and deduplicate viem

SYMBaiEX:
## Summary

- **Fix Privy JWT wallet auth failures**: Flipped token priority in `requirePrivyToken` to prefer the explicit (fresh) token from `getAccessToken()` over the potentially stale `privy-token` HttpOnly cookie. The cookie was passing basic `verifyAuthToken` but failing Privy's stricter wallet authentication with "400 Invalid JWT token provided".
- **Add JWT expiry pre-flight check**: Added a 30-second buffer expiration check in `sendSponsoredEvmTransaction` to catch expired tokens early with a clear error instead of a cryptic Privy API failure.
- **Upgrade `@privy-io/react-auth`** from `^3.8.1` to `^3.13.1` for improved token refresh and cookie management.
- **Upgrade `viem`** from `^2.41.2` to `^2.45.1` across all packages and deduplicate by moving to `peerDependencies` in `@babylon/shared` and `@babylon/api`. This fixes a pre-existing typecheck failure caused by duplicate viem type copies in nested `node_modules`.

## Test plan

- [ ] Verify NFT minting and sponsored transactions succeed without "400 Invalid JWT token" errors
- [ ] Verify wallet funding flow still works end-to-end
- [ ] Confirm `bun run typecheck` passes (all 16 packages)
- [ ] Confirm no regressions in authentication flows (sign in, sign out, token refresh)


SYMBaiEX:
## Fixes Applied

### Addressed Review Feedback
- **Configurable JWT expiry buffer**: `EXPIRY_BUFFER_SECONDS` is now read from `PRIVY_JWT_EXPIRY_BUFFER_SECONDS` env var with a 30s default, allowing adjustment per environment or network latency.
- **Missing appId warning**: When `PRIVY_APP_ID` / `NEXT_PUBLIC_PRIVY_APP_ID` is not set, audience validation is now logged as a warning instead of silently skipped.

### Verification
- All 16 packages pass typecheck
- No linter errors
- Biome check clean

Ready for re-review.

SYMBaiEX:
## Fix Applied: Harden env var parsing

Addressed the automated review feedback on `PRIVY_JWT_EXPIRY_BUFFER_SECONDS` parsing:

- Replaced `Number(env ?? 30)` with `Number.isFinite(parsed) && parsed > 0` guard
- Empty string `""` no longer silently disables the buffer (was coercing to `0`)
- Non-numeric values like `"abc"` no longer bypass the expiry check (was coercing to `NaN`)
- Falls back to `30s` default for any invalid, negative, or missing value

Commit: 53a32c185

PR #1038, opened 2026-02-10
fix: leaderboard total points

slkzgm:
## What
- Restores a scalable Total leaderboard: `pointsType=total` uses DB ordering/pagination on `users.totalPoints` (indexed).
- Fixes totalPoints computation to include positions under both `users.id` and `users.privyId` (aliasing), matching `portfolio-breakdown`.
- Adds a bounded self-heal backfill: `/api/cron/points-recompute` now marks users with `totalPoints=0` as dirty (`TotalPointsService.markZeroTotalPointsDirty`) so the existing 15-min cron can recompute incrementally.
- Ensures new users start with `totalPoints=1000` and are marked dirty (existing change).

## Why
`users.totalPoints` defaults to 0 while `virtualBalance` defaults to 1000. If the recompute cron never ran (or users were never marked dirty), the Total leaderboard showed 0 for everyone.

## How It Heals Existing Data
After deploy, the next `/api/cron/points-recompute` runs will:
- Mark a bounded batch of users with `totalPoints=0` as dirty
- Recompute dirty users (wallet + open perps + open predictions)
- Repeat every 15 minutes until backfilled

## Checks
- `bun run lint --filter=@babylon/api --filter=@babylon/engine --filter=web` OK
- `bun run build --filter=web` OK
- `bun test packages/testing/unit/total-points-service.test.ts` OK
- `turbo typecheck` currently fails due to pre-existing errors in `packages/api` / `packages/training` unrelated to this PR.

### DIFF
--- a/apps/web/src/app/api/cron/points-recompute/route.ts
+++ b/apps/web/src/app/api/cron/points-recompute/route.ts
@@ -45,6 +45,10 @@ export async function POST(request: NextRequest) {
   logger.info('Points recompute started', { isMidnight }, 'PointsRecompute');
 
   try {
+    // Self-heal: mark users with totalPoints=0 as dirty so recompute can backfill.
+    const markedZeroTotalPoints =
+      await TotalPointsService.markZeroTotalPointsDirty();
+
     // Incremental: only recompute users marked dirty
     const recomputeResult = await TotalPointsService.recomputeDirtyUsers();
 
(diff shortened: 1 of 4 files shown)


SYMBaiEX (approved):
LGTM - code changes are correct and well-structured. The DB-level pagination for the total leaderboard is the right approach for scalability, the user aliasing fix aligns with portfolio-breakdown behavior, and the self-heal backfill is bounded and safe. CI failures are pre-existing on staging (typecheck errors in packages/api and packages/training). Minor nice-to-haves for a follow-up: extract the hardcoded '1000' into a shared constant, and remove the now-unreachable 'total' case in the sortField ternary.

PR #1043, opened 2026-02-10
Revert "fix: surface actual NFT mint errors instead of generic 500"

tcm390:
Reverts BabylonSocial/babylon#1042


slkzgm:
Closing: PR #1043 is a revert of #1042 and now conflicts with staging after #1044. Instead of reverting structured mint error handling, we should keep the typed error result and sanitize/mask user-facing messages while retaining server-side diagnostics.

PR #1050, opened 2026-02-10
fix: resolve NFT mint JWT failure and unify auth token sourcing

SYMBaiEX:
## Summary

- **Root cause**: NFT minting failed with `400 Invalid JWT token provided` because Ethereum mainnet (`chainId: 1`) was missing from Privy's `supportedChains` config. All other transaction types (buy/sell shares, on-chain registration) succeeded because they target the app's default chain — only NFT minting targets mainnet.
- **Auth token sourcing**: Unified cookie vs header preference across the codebase — cookies in production, headers in staging/local — controlled by a new `preferCookieAuth()` helper (overridable via `PRIVY_AUTH_PREFER_COOKIE` env var).
- **Diagnostics**: Enhanced `sendSponsoredEvmTransaction` logging with chain context (`caip2`, `chainId`, `sid`, token age) so future chain-specific failures are immediately identifiable.

## Changes

| File | Change |
|------|--------|
| `packages/shared/src/auth/privy-config.ts` | Add `mainnet` to `supportedChains` |
| `packages/api/src/utils/environment.ts` | Add `preferCookieAuth()` helper |
| `packages/api/src/auth-middleware.ts` | Environment-aware token preference in `authenticate`/`optionalAuth` |
| `apps/web/src/app/_actions/utils.ts` | Environment-aware token preference in `requirePrivyTokenBundle` |
| `apps/web/src/app/api/users/onboarding/onchain/route.ts` | Environment-aware token sourcing |
| `packages/api/src/services/privy/evm-send-transaction.ts` | Chain-aware diagnostic logging + final error summary |
| `apps/web/src/hooks/useNftMint.ts` | Auto-retry with fresh token on auth errors |
| `packages/api/package.json` | Align `@privy-io/server-auth` to `^1.32.5` |

## Manual step required

Gas sponsorship for `eip155:1` (Ethereum Mainnet) must be enabled in the **Privy Dashboard** under Gas Sponsorship settings. This is documented in `docs/infra/privy-embedded-wallet-flow.md` but may not have been completed.

## Test plan

- [ ] Verify NFT minting succeeds on mainnet after deploying + enabling gas sponsorship in Privy Dashboard
- [ ] Verify buy/sell shares and on-chain registration still work on the app's default chain
- [ ] Verify auth works correctly on staging (header-preferred) and production (cookie-preferred)
- [ ] Verify the `PRIVY_AUTH_PREFER_COOKIE` env var override works when set explicitly

### DIFF
--- a/apps/web/src/app/_actions/utils.ts
+++ b/apps/web/src/app/_actions/utils.ts
@@ -1,6 +1,5 @@
 'use server';
 
-import { preferCookieAuth } from '@babylon/api';
 import { cookies } from 'next/headers';
 
 type PrivyTokenBundle = {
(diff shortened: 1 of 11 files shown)


slkzgm:
Nice work on this — the summary makes sense and the extra diagnostics are helpful.

A few things I’m a bit worried about before we merge:

- **Token preference (cookie vs header):** switching the primary token by environment via `preferCookieAuth()` might reintroduce the “stale cookie” issue for wallet actions. For Privy wallet endpoints, the safest pattern tends to be “fresh `getAccessToken()` token first, cookie as fallback” rather than cookie-first in prod.
- **Mint retry logic:** the client retry in `useNftMint` is triggered from user-facing error strings (incl. “could not submit the transaction”), which feels too broad and could retry on non-auth failures / make behavior harder to reason about.
- **Dependency bump:** upgrading `@privy-io/server-auth` in the same PR increases blast radius; would be cleaner to split or justify why it’s required.
- **New env var:** if we keep `PRIVY_AUTH_PREFER_COOKIE`, we should document it in `.env.example`.

I think we should keep the good parts (mainnet in `supportedChains`, and the improved logging), but tighten/limit the auth/token selection changes and avoid heuristic retries.

SYMBaiEX:
## Review Feedback Addressed

### Removed
- **`preferCookieAuth()` and all auth token sourcing changes** — reverted `utils.ts`, `auth-middleware.ts`, `onchain/route.ts`, `environment.ts`, `utils/index.ts`, and `index.ts` to their staging originals. Token preference stays as-is: explicit-first for server actions, cookie-first for API routes.
- **Heuristic client retry** — removed `isRetryableAuthError()` and the retry block in `useNftMint.ts`. No more string-matching on user-facing error messages.
- **`@privy-io/server-auth` version bump** — reverted to `^1.18.4` (lockfile already resolves to 1.32.5; spec alignment can go in a separate PR).
- **`PRIVY_AUTH_PREFER_COOKIE` env var** — gone with `preferCookieAuth()`.

### Kept
- **`mainnet` in `supportedChains`** (`privy-config.ts`) — the primary fix
- **Chain-aware diagnostic logging** (`evm-send-transaction.ts`) — `sid`, `tokenAgeSec`, `timeToExpirySec`, `caip2`, `chainId` in pre-flight log; chain context in retry/failure logs; final error summary

### Resulting diff vs staging
Only **2 files** changed: `privy-config.ts` and `evm-send-transaction.ts`.

Ready for re-review.

PR #1059, opened 2026-02-11
fix: sidebar user bar visibility and mobile NFT banner spacing

jasperdisney:
## Summary
- **Desktop sidebar**: Dynamically adjust sidebar height to account for the NFT promo banner, so the user profile bar at the bottom is always visible without scrolling
- **Mobile**: Remove redundant `pt-14` on `<main>` that caused a 56px gap between the NFT banner and page content


SYMBaiEX (approved):
Verified and tested, ready to merge.

PR #1091, opened 2026-02-13
fix: prevent NFT mint database desync with on-chain guards

SYMBaiEX:
## Summary

- Add on-chain `hasMinted()` checks in `checkEligibility()` and `prepareMint()` that auto-reconcile users whose mints succeeded on-chain but were never recorded in the DB (caused by Vercel function timeouts)
- Detect `AlreadyMinted` (`0xddefae28`) revert in `mintNftAction` and trigger reconciliation instead of showing a generic error
- Set `maxDuration=120` on the mint server action to prevent Vercel function timeout during the tx confirmation polling loop

## Root Cause

5 users minted successfully on-chain but the DB was never updated. The Vercel function timeout (default 60s) killed the polling loop before `confirmMint` could verify the transaction receipt and write the DB records. A one-time reconciliation script was already run to fix 4 of 5 affected users; the 5th (Stingface) will be auto-reconciled on next page visit via the new `checkEligibility()` guard. 3 stale records from an old contract deployment were also cleaned up.

## Test plan

- [ ] Verify the 4 reconciled users now see their NFTs on the `/nft` page
- [ ] Verify Stingface sees their NFT after visiting the page (auto-reconciliation via eligibility check)
- [ ] Verify new mints still work end-to-end
- [ ] Verify a user who already minted sees "already minted" status, not a mint button
- [ ] Confirm no regressions for users who haven't minted yet (eligible users still see the mint flow)

### DIFF
--- a/apps/web/src/app/_actions/nft.ts
+++ b/apps/web/src/app/_actions/nft.ts
@@ -1,10 +1,5 @@
 'use server';
 
-// Allow up to 2 minutes for the mint flow (tx submission + confirmation polling).
-// Without this, the default Vercel serverless function timeout (~60s) can kill
-// the polling loop before confirmation completes.
-export const maxDuration = 120;
-
 import {
   extractPrivyApiDiagnostics,
   getAuthedUserContextFromPrivyTokenBundle,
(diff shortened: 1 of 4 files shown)


SYMBaiEX:
@claude review

PR #1093, opened 2026-02-15
feat: move feedback button from floating FAB to sidebar nav

jasperdisney:
## Summary
- Moved the floating feedback button (fixed bottom-right) into the left sidebar as a nav item, eliminating overlap with page content (e.g. leaderboard pagination)
- Styled with a subtle emerald tint to distinguish it as a call-to-action while fitting the sidebar aesthetic
- Aligned the "Feedback" label with the username text in the user menu below
- Deleted the now-unused `FeedbackButton.tsx` floating component


slkzgm (dismissed):
LGTM overall. Moving the Feedback entry from a floating FAB into the Sidebar is a net win (no more overlap with leaderboard/pagination, and it fits the nav IA better).

I pushed a small follow-up fix to improve accessibility of the new Sidebar button (adds aria-label/haspopup/expanded): commit 3f3104fc0.

Notes / minor nits:
- Feedback button: if we care about usage metrics, consider tracking the modal open similar to other sidebar navigation tracking.
- Leaderboard sidebar offset: using container.getBoundingClientRect().top per scroll is OK, but if we see regressions with sticky headers/banners, we may want a more explicit top offset source (CSS var / measured banner height).

Validation I ran locally (worktree):
- bun run typecheck (passed)
- bun run lint (passes, but Biome reports 6 pre-existing info findings in other files)
- apps/web: bun run build (passed with local env present)

No functional concerns with the PR’s main change.

revlentless:
Added metric tracking to the feedback button as @slkzgm suggested. LGTM

PR #1097, opened 2026-02-16
fix: mobile UX — chart dock, chat input, and group leave/delete

jasperdisney:
## Summary
- **BF-72**: Fix bottom dock on markets page hidden behind BottomNav — changed from sticky to fixed positioning with dynamic offset; fixed Markets overlay to use fixed inset-0 for full viewport coverage
- **BF-76**: Simplify chat leave/delete — removed redundant three-dot menu (Leave Chat already in Settings modal); added `onGroupRemoved` callback to fix "Cannot read properties of null" crash after leaving/deleting a group; changed Leave Group button to red
- **BF-78**: Fix chat input behind BottomNav — chat container now uses fixed positioning between MobileHeader and BottomNav; works with existing `interactiveWidget: resizes-content` viewport setting so input moves up when keyboard opens; added safe-area padding for notched devices

### DIFF
--- a/apps/web/src/app/chats/page.tsx
+++ b/apps/web/src/app/chats/page.tsx
@@ -13,17 +13,6 @@ import {
 } from '@/components/chats';
 import { CreateGroupModal } from '@/components/groups/CreateGroupModal';
 import { GroupManagementModal } from '@/components/groups/GroupManagementModal';
-import {
-  AlertDialog,
-  AlertDialogAction,
-  AlertDialogCancel,
-  AlertDialogContent,
-  AlertDialogDescription,
-  AlertDialogFooter,
-  AlertDialogHeader,
-  AlertDialogTitle,
-} from '@/components/ui/alert-dialog';
-import { buttonVariants } from '@/components/ui/button';
 import { useA2A } from '@/hooks/useA2A';
 import { useAuth } from '@/hooks/useAuth';
 import { useChatParam } from '@/hooks/useChatParam';
@@ -75,14 +64,6 @@ export default function ChatsPage() {
     sendError,
     sendWarning,
     sendSuccess,
-
-    // Leave chat
-    isLeaveConfirmOpen,
-    setLeaveConfirmOpen,
-    isLeavingChat,
-    leaveChatError,
-    setLeaveChatError,
-    handleLeaveChat,
 
     // Group modals
     isCreateGroupModalOpen,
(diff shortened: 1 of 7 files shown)


revlentless:
Reviewed all changes across the 7 files. Overall this is solid — three real bugs fixed cleanly. Left a follow-up commit addressing the minor issues I found.

---

### BF-72: Bottom dock hidden behind BottomNav ✅

Switching the mobile bottom dock from `sticky` to `fixed` + using the full measured `mobileBottomNavHeight` as the offset is the right approach. The `ResizeObserver` on `#app-bottom-nav` ensures this stays correct across devices with different safe-area insets.

The markets overlay also needed `fixed inset-0` instead of `absolute inset-0` — good catch, that would've been clipped by any non-static ancestor.

### BF-76: Chat leave/delete simplification ✅

Clean removal of the redundant three-dot menu. Leave Group now lives exclusively in the Settings modal which is the right UX — one path to the action, not two.

The `onGroupRemoved` callback is the key fix here. Previously, leaving/deleting a group called `onGroupUpdated` which just refreshed the chat list but left the selection pointing at a now-deleted group → null crash. The new callback correctly clears `selectedChatId`, `selectedGroupId`, closes the modal, and reloads the chat list.

Yellow → red for the Leave Group button makes sense from a destructive-action UX perspective.

### BF-78: Chat input behind BottomNav ✅

`fixed inset-x-0 top-14 bottom-14` with `md:relative md:inset-auto` is a clean solution. Avoids the fragile `calc(100dvh - 112px)` approach and works naturally with `interactiveWidget: resizes-content` for keyboard handling.

The `pb-safe md:pb-0` addition on the chat footer handles notched devices properly.

> **Note:** `top-14` and `bottom-14` are hardcoded to match MobileHeader and BottomNav heights. If those ever change, this will need updating too. Might be worth a comment noting the coupling.

---

### Follow-up commit pushed

Addressed two minor issues in `a645d7ccb`:

1. **Removed dead leave-chat code** — The `AlertDialog`, its state (`isLeaveConfirmOpen`, `isLeavingChat`, `leaveChatError`, etc.), and the `AlertDialog`/`buttonVariants` imports in `chats/page.tsx` were no longer reachable after `onLeaveChat` was removed. Cleaned all of it up (-58 lines).

2. **Fixed redundant Tailwind class** — `px-3 p-3` in `MarketsTradingTerminal.tsx` — `p-3` already covers horizontal padding, so `px-3` was a no-op. Changed to just `p-3`.

---

### Nit (no action needed)
- make sure you run `bun run check` to run biome formatting before you PR


LGTM!

PR #1103, opened 2026-02-17
ci: make fork PR workflows resilient to missing secrets

lawyered0:
## Summary
- add fork-safe fallback handling for Privy app id env in CI setup/build workflows
- gate Synpress and Playwright E2E steps behind credential-availability flags
- skip Claude code review workflow on fork PRs where required auth context is unavailable

## Why
Fork pull requests do not receive repository secrets, which caused CI jobs to fail before evaluating code changes. This makes CI behavior reflect workflow constraints rather than code quality.

## Validation
- `bun run validate:all` ✅
- workflow file changes only (no runtime app/test logic changes)

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -117,9 +117,10 @@ jobs:
           # Use CI test database directly
           DATABASE_URL: postgresql://postgres:postgres@localhost:5432/test_db
           DIRECT_DATABASE_URL: postgresql://postgres:postgres@localhost:5432/test_db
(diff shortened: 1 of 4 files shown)


revlentless:
Nice and clean.

The `HAS_*_SECRETS` env flag pattern in `ci-tests.yml` for gating Synpress/E2E is a good approach. One thing I noticed: `ci-unit-integration.yml` has integration + Playwright auth steps that also rely on Privy creds but aren't gated the same way. Not a blocker since those steps already have `continue-on-error` style fallbacks, but might be worth a quick follow-up to keep things consistent.

Also minor — the dummy app id `cl_ci_fork_dummy` works fine, just might look confusing in logs if someone's debugging a fork PR build. Something like `FORK_PR_PLACEHOLDER` could be clearer, but that's purely cosmetic.

LGTM 👍

PR #1107, opened 2026-02-17
test: move DB-backed NPC PnL spec from unit to integration

lawyered0:
## Summary
- move `packages/testing/unit/npc-investment-manager-pnl.test.ts` to `packages/testing/integration/npc-investment-manager-pnl.integration.test.ts`
- keep test logic unchanged (real DB read/write behavior remains)
- update header comment to explicitly mark this as an integration test

## Why
- this spec performs real DB operations and fails unit runs when `DATABASE_URL` is not configured
- keeping it in `unit` introduces a false infrastructure dependency in the unit gate

## Validation
- `bun run --cwd packages/testing typecheck`
- `bun run --cwd packages/testing lint`
- `bun run validate:all`
- integration test requires DB env (expected to fail locally without `DATABASE_URL`)

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -19,9 +19,6 @@ jobs:
     name: Run All Tests
     runs-on: ubuntu-latest
     timeout-minutes: 60
-    env:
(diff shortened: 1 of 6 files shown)


revlentless:
Moving the PnL test to `integration/` makes sense. It hits a real DB so it never belonged in the unit gate. The mock refactor for `AutomationPipeline.test.ts` is a nice cleanup too; spreading the actual module instead of hand-listing every export is way less brittle.

LGTM!

PR #1134, opened 2026-02-22
fix: resolve typecheck failures in onchain-service

revlentless:
## Summary

- Fix two TypeScript errors in `onchain-service.ts`: mismatched `awardPoints` parameter type (`string` vs `PointsReason`) and a null-safety issue on the fallback return path.
- Fix `web` package typecheck consistently crashing with OOM by increasing the Node.js heap limit (matching the `build` script).

## Type

- [x] Bug fix
- [x] Infra / ops

## Context / Links

- `bun run typecheck` was failing across the monorepo due to these two independent issues.

## Scope (keep it focused)

- [x] This PR is focused on a single change/theme (not a catch‑all)
- [x] Drive‑by refactors are excluded or split into a separate PR
- [x] Non‑goals / follow‑ups are listed below (with links)

**Areas touched**
- [x] Web UI (`apps/web`)

## Changes

- `packages/api/src/services/onchain-service.ts`: Changed `OnboardingServices.awardPoints` parameter type from `string` to `PointsReason` to match `PointsService.awardPoints` signature. Introduced a typed local variable in `getOnboardingServices()` to avoid returning a value still typed as `| null`.
- `apps/web/package.json`: Added `NODE_OPTIONS='--max-old-space-size=8192'` to the `typecheck` script (already present on the `build` script).
- `apps/web/src/components/shared/ComingSoon.tsx`, `comingSoonAccess.test.ts`: Import sort order and formatting (Biome auto-fix).

## Review guide

**Start here**
- `packages/api/src/services/onchain-service.ts`

**Risk**
- [x] Low

**Notes for reviewers**
- The `PointsReason` type change is safe — `PointsService.awardPoints` already required this type; the `OnboardingServices` interface was simply lagging behind.
- The OOM fix mirrors the existing `build` script approach.

## Test plan

**Commands run**
- [x] `bun run typecheck`

**Manual verification**
1. `bun run typecheck` now passes all 16 packages (previously failed on `@babylon/api` and `web`).

- [x] No deploy impact

## Breaking changes

- [x] None

## Security / privacy

- [x] No security impact

- Reason: Type-level fix and build config change, no UI impact.

## Checklist (author)

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -11,7 +11,7 @@
     "lint": "biome lint .",
     "lint:fix": "biome check --write .",
     "format": "biome format --write .",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "NODE_OPTIONS='--max-old-space-size=8192' tsc --noEmit",
     "clean": "rm -rf .next .turbo node_modules/.cache"
   },
   "dependencies": {
--- a/apps/web/src/components/shared/ComingSoon.tsx
+++ b/apps/web/src/components/shared/ComingSoon.tsx
@@ -28,13 +28,13 @@ import { createPortal } from 'react-dom';
 import { toast } from 'sonner';
 import { LinkSocialAccountsModal } from '@/components/profile/LinkSocialAccountsModal';
 import { Avatar } from '@/components/shared/Avatar';
-import { MarketingFooter } from '@/components/shared/MarketingFooter';
-import { PlayerStatsModal } from '@/components/shared/PlayerStatsModal';
 import {
   getPrimaryAccessLabel,
-  shouldAutoRedirectWhitelistedUser,
   type NftAccessState,
+  shouldAutoRedirectWhitelistedUser,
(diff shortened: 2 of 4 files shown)


slkzgm (approved):
Looks good.

- `packages/api/src/services/onchain-service.ts`: Aligning `OnboardingServices.pointsService.awardPoints` to `PointsReason` and returning a typed local `services` value fixes the strict function-assignability + nullability issues cleanly.
- `apps/web/package.json`: Adding `NODE_OPTIONS=--max-old-space-size=8192` to the web `typecheck` script matches the existing `build` script and avoids OOM during `tsc --noEmit`.
- `ComingSoon.tsx` / `comingSoonAccess.test.ts`: Biome-only import/formatting changes; no behavior change.

Verified on this branch:
- `bun run typecheck`
- `bun run lint`

Non-blocking: if we ever need cross-shell portability (e.g. Windows), we may want to switch env var setting to a cross-platform approach, but current pattern is consistent with `build`.

PR #1135, opened 2026-02-22
testing: fix agent-tick-db unit crash from server-only import ordering

lawyered0:
## Summary
- load the `agent-tick` route via top-level dynamic import so test mocks are registered before module evaluation
- mock `server-only` and `@/lib/engine/ensure-engine-services` in the unit test to prevent server-component guards from crashing Bun unit runs
- preserve full `@babylon/db` export compatibility by spreading actual module exports and overriding only test-specific members
- add `afterAll(() => mock.restore())` to avoid cross-suite mock leakage

## Problem
`agent-tick-db.test.ts` could fail before executing any tests due to module-load order and incomplete mocked exports.

## Validation
- `bun test packages/testing/unit/cron/agent-tick-db.test.ts --preload ./packages/testing/unit/preload.ts` (pass)
- `bun test packages/testing/unit/cron/article-tick.test.ts packages/testing/unit/cron/markets-tick.test.ts packages/testing/unit/cron/agent-tick-db.test.ts --preload ./packages/testing/unit/preload.ts` (pass)
- `bun run --cwd packages/testing lint` (pass)
- `bun run --cwd packages/testing typecheck` fails only on existing baseline `onchain-service.ts` errors (addressed separately in #1134)

### DIFF
--- a/packages/testing/unit/cron/agent-tick-db.test.ts
+++ b/packages/testing/unit/cron/agent-tick-db.test.ts
@@ -1,5 +1,6 @@
-import { beforeEach, describe, expect, mock, test } from 'bun:test';
+import { afterAll, beforeEach, describe, expect, mock, test } from 'bun:test';
 import { NextRequest } from 'next/server';
+import * as actualDbModule from '../../../db/src/index';
 
 /**
  * Mock game state type
@@ -50,6 +51,8 @@ interface SqlCondition {
 // Mock db with a mutable state we can control in tests
 let mockGame: MockGame | null = null;
 
+mock.module('server-only', () => ({}));
+
 // Create a complete mock that includes schema exports
 mock.module('@babylon/db', () => {
   const createModelMock = (overrides: Partial<MockModel> = {}): MockModel => ({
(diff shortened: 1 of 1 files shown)


revlentless:
Reviewed and tested locally. LGTM.

**What I verified:**

- `bun test packages/testing/unit/cron/agent-tick-db.test.ts --preload ./packages/testing/unit/preload.ts` - 3/3 pass
- All three cron suites together (`article-tick`, `markets-tick`, `agent-tick-db`) - 84/84 pass, no cross-suite mock leakage

**Review notes:**

- The `await import()` switch, `server-only` mock, and `ensure-engine-services` mock all make sense - they fix the module-evaluation-before-mocks-registered ordering issue.
- Spreading `actualDbModule` into the `@babylon/db` mock brings this test in line with the pattern already used in `markets-tick.test.ts` and `article-tick.test.ts`. Good consistency fix - also makes the test resilient to new exports being added to `@babylon/db`.
- `afterAll(() => mock.restore())` matches the other cron test files. Clean.

**One optional suggestion for a follow-up (not blocking):**

Since `server-only` is a Next.js server-component guard that will never work in Bun test runs, it might be worth mocking it globally in `preload.ts` rather than per-test-file. Right now this is the only test that needs it, but if more route tests land that transitively import it, centralizing the mock would prevent the same crash from recurring elsewhere.

PR #1137, opened 2026-02-22
testing: align formatCurrency negative sign expectations

lawyered0:
## Summary
- update `packages/testing/unit/shared/format.test.ts` negative-currency expectations to match canonical formatter output (`ƀ-...`)
- keep assertion behavior consistent with existing shared currency tests and runtime formatter semantics

## Problem
`format.test.ts` expected negative values as `-ƀ...`, but the shared formatter intentionally returns `ƀ-...`. This caused deterministic unit failures in the full suite.

## Validation
- `bun test packages/testing/unit/shared/format.test.ts --preload ./packages/testing/unit/preload.ts` (pass)
- `bun test packages/testing/unit/shared/format-currency.test.ts --preload ./packages/testing/unit/preload.ts` (pass)

### DIFF
--- a/packages/testing/unit/shared/format.test.ts
+++ b/packages/testing/unit/shared/format.test.ts
@@ -150,9 +150,9 @@ describe('Format Utilities', () => {
     });
 
     it('should handle negative numbers correctly', () => {
-      // Sign should come before the symbol
-      expect(formatCurrency(-100)).toBe('-ƀ100.00');
-      expect(formatCurrency(-1234.56)).toBe('-ƀ1234.56');
+      // Shared formatter keeps symbol first for positive and negative values.
+      expect(formatCurrency(-100)).toBe('ƀ-100.00');
+      expect(formatCurrency(-1234.56)).toBe('ƀ-1234.56');
     });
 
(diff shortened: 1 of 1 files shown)


revlentless:
LGTM.

**What I verified:**

- `bun test packages/testing/unit/shared/format.test.ts --preload ./packages/testing/unit/preload.ts` - 36/36 pass
- `bun test packages/testing/unit/shared/format-currency.test.ts --preload ./packages/testing/unit/preload.ts` - 40/40 pass

**Review notes:**

Confirmed this against the actual `formatCurrency` implementation in `packages/shared/src/utils/format.ts`. The formatter explicitly handles negatives as symbol-first, then sign:

```ts
const sign = isNegative ? '-' : '';
return `${BABYLON_POINTS_SYMBOL}${sign}${formatted}`;
```

So `ƀ-100.00` is the correct output - the tests in `format.test.ts` were simply wrong (asserting `-ƀ100.00`). The sibling test file `format-currency.test.ts` already had the right expectations (`ƀ-100.00`), so this brings the two into alignment.

PR #1138, opened 2026-02-23
fix(ui): align chat send button to bottom of input area

jasperdisney:
## Summary

Fixes the chat message input send button alignment in `/chats` and `/agents/team` pages. Previously, when the textarea grew to multiple rows, the send button would vertically center itself in the expanding input area. Now it stays anchored at the bottom-right.

- Changed composer shell from `items-center` to `items-end` so the send button stays at the bottom as the textarea grows
- Adjusted container padding (`pl-3 pr-1 pb-1`) and removed symmetric `py-2` to give the button a tight bottom-right position
- Changed textarea padding from `py-[5px]` to `pt-2` (top only) so text aligns naturally while bottom spacing is handled by the container
- Resized send button from `h-8 w-8` to `size-[34px]`
- Updated highlight overlay padding to match (`pt-2`)

Both `/chats` and `/agents/team` share the same `MessageInput` component, so the fix applies to both pages.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -11,7 +11,7 @@
     "lint": "biome lint .",
     "lint:fix": "biome check --write .",
     "format": "biome format --write .",
-    "typecheck": "tsc --noEmit",
+    "typecheck": "NODE_OPTIONS='--max-old-space-size=8192' tsc --noEmit",
     "clean": "rm -rf .next .turbo node_modules/.cache"
   },
   "dependencies": {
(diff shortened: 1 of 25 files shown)


revlentless:
Looks good. The core fix (`items-center` → `items-end`) is the right call - send button now stays anchored at the bottom-right as the textarea grows instead of floating to the center.

A few notes on the final state:

- Nice cleanup in the refactor commit swapping arbitrary values (`size-[34px]`, `leading-[22px]`, `min-h-[32px]`) for built-in Tailwind classes (`size-8.5`, `leading-5.5`, `min-h-8`). Much cleaner.
- Textarea has `pt-2` with no bottom padding - bottom spacing is handled by the container's `pb-1` (4px). Looks intentional but worth a quick visual check on multi-line messages to make sure the last line of text doesn't feel too tight against the border.
- The highlight overlay and both textarea paths (mentions enabled/disabled) all use matching `pt-2` + `leading-5.5`, so mention highlights should stay aligned with the underlying text.

Approved

PR #1139, opened 2026-02-23
feat(ui): polish chats, rewards, and group modals for mobile

jasperdisney:
## Summary

- **Chat back button**: Simplified to icon-only (removed "Back" text), larger arrow
- **Chat header**: Removed group icon, reduced height on mobile, moved SSE Live badge to participants row for more title space, tightened gap between title and subtitle
- **Chat sidebar**: Removed inner shadow from group icons, reduced whitespace before search/filter tabs
- **Message loading skeleton**: Updated to match real chat bubble layout (avatars, sender names, alternating left/right bubbles)
- **Rewards page**: Top 3 stats stack vertically on mobile (Total Points first), Earned & Referrals side-by-side, removed separator between Earn Points and Referral Link sections
- **Group settings modal**: Type badge on separate row below name, name truncates with ellipsis, removed gear icon, reduced padding and spacing
- **Create group modal**: Character count inline with Group Name label (right-aligned), no layout shift on input

### DIFF
--- a/apps/web/src/app/globals.css
+++ b/apps/web/src/app/globals.css
@@ -369,15 +369,6 @@ button:disabled {
   box-shadow: inset 5px 5px 5px rgba(0, 0, 0, 0.1), inset -5px -5px 5px rgba(255, 255, 255, 0.05);
 }
 
-.chat-button {
-  box-shadow: inset 3px 3px 3px rgba(0, 0, 0, 0.1), inset -3px -3px 3px rgba(255, 255, 255, 0.05);
-  transition: all 0.3s ease;
-}
-
-.chat-button:hover:not(:disabled) {
-  box-shadow: none;
-}
-
 .message-input {
   box-shadow: none;
 }
@@ -398,10 +389,6 @@ button:disabled {
 /* Dark mode adjustments for chat styles */
 .dark .chat-card {
   box-shadow: inset 5px 5px 5px rgba(0, 0, 0, 0.3), inset -5px -5px 5px rgba(255, 255, 255, 0.02);
-}
-
-.dark .chat-button {
-  box-shadow: inset 3px 3px 3px rgba(0, 0, 0, 0.3), inset -3px -3px 3px rgba(255, 255, 255, 0.02);
 }
 
 .dark .message-input {
--- a/apps/web/src/app/rewards/page.tsx
+++ b/apps/web/src/app/rewards/page.tsx
@@ -664,20 +664,7 @@ export default function RewardsPage() {
             </div>
 
             {/* Stats Row */}
-            <div className="grid grid-cols-3 gap-2 sm:gap-3">
-              {/* Total Earned */}
-              <div className="rounded-lg border border-border p-3">
-                <div className="mb-1 flex items-center gap-1">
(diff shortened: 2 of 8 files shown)


revlentless:
Good set of polish changes. Reviewed the diff across all 7 files - everything looks clean and intentional. Notes below.

**Chat header (`ChatViewHeader.tsx`)**
- Back button simplified to icon-only with `aria-label="Back"` - good for accessibility. Larger arrow (`h-5 w-5`) is easier to tap on mobile.
- Removing the group `<Users>` icon from the header is fine - the group name + participant count is enough context, and it frees up horizontal space. The `Users` import was cleaned up properly.
- SSE Live badge moved to the same row as the participant count with `-mt-0.5` on mobile. Works well - tightens the header without losing information. For DMs (no participant count), the SSE badge sits alone in that row, which is consistent.
- Mobile padding reduced from `py-4` to `py-2 md:py-4` - good density improvement.

**Chat sidebar (`ChatHeader.tsx`, `ChatListItem.tsx`)**
- `mb-4` → `mb-2` before filter tabs and removing `mb-4` from the tab container - tighter layout, less dead space before search. Good.
- `chat-button` class removed from the group icon in `ChatListItem.tsx`. **This was the last usage** - the `.chat-button` CSS rules in `globals.css` (lines 372–405) are now dead code. Should be cleaned up in this PR or a follow-up.

**Message loading skeleton (`MessageList.tsx`)**
- Much better than the old three stacked rectangles. Alternating left/right bubbles with avatars and sender name placeholders actually looks like a chat. The `flex-1 flex-col justify-end` anchors the skeletons to the bottom like real messages would appear. Nice touch.

**Rewards page (`rewards/page.tsx`)**
- The `space-y-2 sm:grid sm:grid-cols-3 sm:gap-3 sm:space-y-0` + `grid grid-cols-2 gap-2 sm:contents` pattern is a clean way to stack Total Points full-width on mobile while keeping Earned and Referrals side-by-side. On sm+ it all flattens back to a 3-column grid via `sm:contents`. Smart approach.
- Reordering to put Total Points first makes sense - it's the primary metric.
- One `<Separator />` removed between Earn Points and Referral Link. The other `<Separator />` at line 511 is still used, so the import stays. Fine.

**Group settings modal (`GroupManagementModal.tsx`)**
- `truncate` on the group name heading + `min-w-0 flex-1` on the container - proper truncation setup for long group names. Good.
- Type badge on a separate row (`mt-1`) gives the name full width. Better than cramming both on one line.
- Settings gear icon removed from the modal header. Since you're already inside the settings modal, it was redundant. Clean.
- Header padding `p-6` → `px-6 py-4`, content `p-6` → `px-6 pt-4 pb-6`. Tighter top, same sides. Good.
- Removed `mt-2` from the group name display text (line 723) - consistent with the reduced spacing.

**Create group modal (`CreateGroupModal.tsx`)**
- Character count moved inline next to the label with `flex items-baseline justify-between`. No more layout shift from the count appearing below the input. Clean.

Dead `.chat-button` CSS in `globals.css` was cleaned up in a follow-up commit (`5a0d912`).

Approved.

PR #1142, opened 2026-02-26
feat(mobile): improve chat keyboard UX and hide nav on keyboard open

jasperdisney:
## Summary
- Hide BottomPanel on mobile (keep on desktop) to maximize chat message area
- Hide bottom nav bar when virtual keyboard opens using `visualViewport` API, with `--bottom-nav-height` CSS variable so page layouts reclaim the space without leaving a white gap
- Add safe-area padding (`pb-safe`) to TeamChatView footer for notched devices
- Scroll chat to bottom on input focus (keyboard open) with 150ms delay for smooth behavior
- Fix scroll-to-bottom on initial load for mobile by finding the visible scroll container instead of relying on a shared ref that points to the hidden desktop view

### DIFF
--- a/apps/web/src/app/agents/loading.tsx
+++ b/apps/web/src/app/agents/loading.tsx
@@ -2,7 +2,7 @@ import { Skeleton } from '@/components/shared/Skeleton';
 
 export default function AgentsLoading() {
   return (
-    <div className="flex h-[calc(100dvh-112px)] flex-col md:h-dvh">
+    <div className="flex h-[calc(100dvh-56px-var(--bottom-nav-height))] flex-col md:h-dvh">
       <div className="flex min-h-0 flex-1 overflow-hidden">
         {/* Member sidebar skeleton */}
         <div className="hidden w-64 flex-col border-border border-r p-4 lg:flex">
--- a/apps/web/src/app/agents/team/page.tsx
+++ b/apps/web/src/app/agents/team/page.tsx
@@ -627,13 +627,19 @@ export default function TeamChatPage() {
     // Only scroll when loaded data
     if (!teamChat?.chatId || loading) return;
 
-    const endMarker = messagesEndRef.current;
-    if (!endMarker) return;
-
-    // Find the scroll container
-    const container = endMarker.closest('[data-chat-messages-container]');
+    // Find the visible scroll container (mobile and desktop render separate
+    // TeamChatView instances — pick the one that's actually visible).
+    const containers = document.querySelectorAll<HTMLElement>(
+      '[data-chat-messages-container]'
+    );
+    let container: HTMLElement | null = null;
+    for (const el of containers) {
+      if (el.offsetHeight > 0) {
+        container = el;
+        break;
+      }
+    }
     if (!container) {
-      // Fallback: just scroll once
       scrollToBottom('instant');
       return;
     }
@@ -646,7 +652,7 @@ export default function TeamChatPage() {
     let isActive = true;
 
     const scrollToEnd = () => {
-      endMarker.scrollIntoView({ behavior: 'auto', block: 'end' });
+      container.scrollTop = container.scrollHeight;
     };
 
     const finish = () => {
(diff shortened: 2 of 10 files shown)


revlentless:
Solid approach to a tricky mobile problem. The `--bottom-nav-height` CSS variable is the right abstraction — dynamic height calculations that respond to keyboard state without hardcoded pixel math. Notes below.

**`--bottom-nav-height` CSS variable (`globals.css`, `layout.tsx`, `BottomNav.tsx`)**
- Good pattern: `:root` defaults to `56px`, the `visualViewport` resize handler toggles it to `0px` when keyboard opens, and `layout.tsx` uses `pb-[--bottom-nav-height]` instead of the old hardcoded `pb-14`. This means every page gets the dynamic behavior for free.
- The 150px threshold for keyboard detection is reasonable — avoids false positives from mobile browser chrome changes (URL bar collapsing, etc.).
- One edge case: `fullHeight` is initialized from `vv.height` at mount time and only updated upward. If the component somehow mounts while the keyboard is already open (unlikely given it's in the root layout), `fullHeight` would be wrong. Not a real concern in practice.
- Minor: the comment on line 73 still references the old `h-[calc(100dvh-112px)]` pattern, but that pattern no longer exists anywhere — it's been replaced with `100dvh-56px-var(--bottom-nav-height)`. Stale comment.

**Height calculations (`team/page.tsx`, `loading.tsx`, `markets/page.tsx`, `agents/loading.tsx`)**
- All updated from hardcoded `100dvh-112px` to `100dvh-56px-var(--bottom-nav-height)`. The `56px` is the top header, the variable handles the bottom nav. On desktop, the `md:h-dvh` override takes over, so the variable is irrelevant there. Clean.

**BottomPanel hidden on mobile (`team/page.tsx`)**
- `<div className="hidden lg:contents">` wrapping BottomPanel is a clean pattern — `display: contents` on desktop makes the wrapper invisible to the layout, while `hidden` on mobile removes it entirely. Good use of `contents`.

**Scroll-to-bottom on focus (`MessageInput.tsx`, `TeamChatView.tsx`, `team/page.tsx`)**
- `onInputFocus` threaded through cleanly: `MessageInput` → `TeamChatView` → `team/page.tsx`.
- The `setTimeout(() => scrollToBottom('smooth'), 150)` delay accounts for the keyboard animation. 150ms is a good middle ground — long enough for the viewport to start resizing, short enough to feel responsive.
- **Note**: this only applies to `TeamChatView` (agent team chat). The DM `ChatView` (`ChatView.tsx`) already has `pb-safe md:pb-0` but doesn't get the `onInputFocus` prop. If the keyboard UX fix matters for team chat, it likely matters for DM chat too. Worth adding in this PR or a follow-up.

**Scroll container resolution (`useTeamChat.ts`, `team/page.tsx`)**
- The `getScrollContainer` fix in `useTeamChat.ts` is the right approach — checking `offsetHeight > 0` to find the visible container when two `TeamChatView` instances exist (mobile/desktop).
- The same logic is **duplicated** in the initial scroll effect in `team/page.tsx` (lines 631–642) — it does its own `querySelectorAll` + `offsetHeight > 0` instead of calling `getScrollContainer` from the hook. Consider refactoring to reuse `getScrollContainer` to avoid the duplication.
- `messagesEndRef` was correctly removed from the effect's dependency array since it's no longer used inside the effect body.

**Safe-area padding (`TeamChatView.tsx`)**
- `pb-safe lg:pb-0` on the footer. Note that `ChatView.tsx` (DM chat) uses `pb-safe md:pb-0` — the breakpoints differ (`lg` vs `md`). This is probably intentional given the different layout structures, but worth confirming.

**Things to address:**
1. The `onInputFocus` scroll-on-keyboard-open behavior is missing from `ChatView.tsx` (DM chats) — should it be added?
2. The scroll container lookup in `team/page.tsx`'s initial scroll effect duplicates logic from `getScrollContainer` in `useTeamChat.ts` — consider reusing it.
3. Stale comment in `BottomNav.tsx` line 73 referencing `h-[calc(100dvh-112px)]`.

Approved. (with the notes above as optional follow-ups or in this PR).

PR #1150, opened 2026-02-28
fix: mobile viewport height and tone down trading card colors

jasperdisney:
## Summary
- Replace `min-h-screen` (100vh) with `min-h-dvh` on mobile across all pages to eliminate extra scrollable space caused by mobile browser chrome (address bar/bottom UI)
- Mute NPC/agent badge colors from purple to gray in trade cards, position cards
- Remove colored text from liquidation price and funding paid in perp position cards for a less visually noisy UI
- Fix type error in game-guide test (direct tuple index access)


revlentless:
LGTM!

PR #1163, opened 2026-03-03
feat(feed): narrative feed API — story scoring, ArcState multiplier, resolution boost & position signal

SYMBaiEX:
## Summary

Full implementation of the `/api/feed/narrative` route across three phases. This is the first time any narrative feed code is being pushed to remote.

### Phase 1 — Narrative feed foundation (`route.ts`)
- Introduced `/api/feed/narrative` API route: groups posts into stories by `storyKey`, ranks them with a multi-signal scoring formula
- **Scoring formula**: `storyScore = totalEngagement * 0.5 + recencyScore * 0.35 + activityBonus * 0.15`
  - `totalEngagement = likes*1 + comments*2 + shares*3`
  - `recencyScore = exp(-ln2 * hoursOld / 12)` — 12-hour half-life on the newest post in the story
  - `activityBonus = min(postCount / 10, 1.0)`
- Stories bucketed into question-linked vs general; general story always sorted last
- Per-post `isLiked` / `isShared` enrichment for authenticated users
- Redis caching with 60-second TTL

### Phase 2 — CTE refactor + ArcState multiplier
- Replaced 3 parallel `COUNT(*)` queries with a single engagement CTE (mirrors the `fetchPostMetadataConsolidated` pattern in `posts/route.ts`)
- Added `calculateArcStateMultiplier` — story scores multiplied by arc state (crisis 1.4×, climax 1.35×, revelation 1.3×, escalation 1.15×, active/live 1.1×, tension 1.05×, resolving 0.9×, resolution 0.85×)
- `LEFT JOIN arcStates` on the questions query to fetch title + arc state + status in one round-trip
- Filtered resolved/expired questions from the feed
- Added `arcState` field to `NarrativeStory` interface for UI badge support
- 19 unit tests + 32 integration tests

### Phase 3 — Resolution proximity boost + per-user position signal
- **`calculateResolutionBoost`**: Multiplies story scores based on how soon `questions.resolutionDate` occurs (1.4× ≤6h, 1.25× ≤24h, 1.1× ≤72h, 1.0× otherwise/past) — questions at peak uncertainty float higher
- **Per-user position signal**: `hasUserPosition: boolean` on `NarrativeStory`. Authenticated users with an active position in the `positions` table see those stories floated above non-position stories (within non-general tier), then by score. General story always last.
- Combined `Promise.all([reactions, shares, positions])` in enrichment block — no extra DB round-trip
- Re-sort runs only inside the `if (userId)` block; unauthenticated users get score-only ordering
- Expanded to 24 unit tests + 40 integration tests

### Pre-existing TypeScript fixes (unblocked push hook)
- `packages/testing`: Added `!` non-null assertion on `lastSlide` in `game-guide.test.ts` (`TS18048`)
- `packages/examples/babylon-typescript-agent`: Fixed `registerIPFS()` to call `await handle.waitMined()` before accessing `result.agentId` / `result.agentURI`

## Files Changed

| File | Description |
|------|-------------|
| `apps/web/src/app/api/feed/narrative/route.ts` | Full narrative feed implementation (all phases) |
| `apps/web/src/app/api/feed/narrative/route.test.ts` | 24 unit tests (scoring functions + arc state + resolution boost) |
| `apps/web/src/app/api/feed/narrative/route.integration.test.ts` | 40 integration tests (contract, caching, auth enrichment, arc state, resolution boost, position signal) |
| `packages/testing/unit/game-guide.test.ts` | Non-null assertion fix |
| `packages/examples/babylon-typescript-agent/src/registration.ts` | `waitMined()` fix |

## Test Plan

- [ ] `bun test ./apps/web/src/app/api/feed/narrative/` → **64 pass** (24 unit + 40 integration)
- [ ] `GET /api/feed/narrative` response includes `arcState` and `hasUserPosition: boolean` on every story
- [ ] Stories with active arc states (crisis, climax) rank above equivalent setup/neutral stories
- [ ] Story with `resolutionDate` within 6 hours scores higher than identical story resolving in 30 days
- [ ] Authenticated user with a position on question #N sees that story ranked first within its tier
- [ ] General (no-question) story always appears last regardless of arc state, resolution date, or user positions
- [ ] Unauthenticated requests return `hasUserPosition: false` on all stories, no positions query fired


slkzgm (approved):
Overall this looks solid: clear scoring model, good caching, and strong test coverage (unit + integration). CI is green.

I pushed a small follow-up commit on this branch to align behavior with the PR spec and reduce duplication:
- Filter per-user position signal to *active* positions only (avoids floating stories based on closed/liquidated history).
- Extract scoring helpers into a shared  module and import them in  (so tests cover the production implementation without redefining the formula).

Commit: 744b33f6e (fix(feed): use active positions + centralize narrative scoring)

Non-blocking follow-ups / questions:
- PR description mentions filtering “expired” questions. Current behavior only skips  and keeps past  stories (with 1.0 boost). Is that intended?
- Consider moving more of the grouping/scoring logic out of the Next route (apps thin) into a package/service once we stabilize the API surface.
- If we expect feed growth, adding pagination (/) would avoid returning an ever-growing list of stories.

PR #1181, opened 2026-03-04
feat(feed): Stories UI overhaul — flat interleaved feed, market card redesign, chronological Latest

SYMBaiEX:
## Summary

Visual and UX overhaul of the Stories feed and New Market cards.

### Stories tab — flat jumbled feed
- Delete NarrativeStoryCard — all story header chrome (BookOpen icon, market title, arc-state badges, post count, expand/collapse) removed
- NarrativeStoryList rewritten with round-robin interleaving — 1 post per story per pass, so posts from different markets are jumbled rather than grouped in blocks
- New Market entries appear at their scored position in the flat list
- Posts render at default density as plain feed cards

### NewMarketCard redesign
- Remove "New Market" pill badge
- YES / NO probability bars (50/50 default — new markets open at parity)
- Resolution countdown + "Trade YES" / "Trade NO" split buttons + "View market →"

### Latest tab — chronological market card injection
- New MixedFeedList merges FeedPost[] + NewMarketEntry[] sorted newest-first by timestamp
- Markets appear at correct chronological position (not batched at top)
- All PostList behaviors preserved (infinite scroll, invite banner, etc.)

### Scope restriction
- useNewMarkets only enabled for Latest tab (was also Hot)
- Hot, Following, Trades tabs: no market cards at all

## Test Plan
- Stories: posts from different markets interleaved (not grouped under headers)
- Stories: no BookOpen, no market titles, no arc-state badges
- Latest: new market cards at correct chronological position
- Hot/Following: no market cards
- NewMarketCard: YES/NO bars, countdown, Trade YES/NO buttons


slkzgm (dismissed):
## Review summary\n\nThis PR delivers the described UX changes:\n- **Stories**: flat interleaved feed (no story header chrome) + new-market cards in the scored stream.\n- **Latest**: posts + new markets merged chronologically (no longer pinned/batched).\n- **Other tabs**: no market cards.\n\n### What I checked\n- Algorithms are extracted into pure utilities and are covered by unit tests.\n- Client rendering paths are memoized () where it matters.\n- API routes return  + live share counts to support deep links and probability bars.\n\n### Non-blocking concerns / follow-ups\n- **DB perf**: joining  via  will not use a normal index. Given caching this is probably OK short-term, but we should migrate to a real FK (or add a functional index) as a follow-up.\n- **Type boundaries**: consider moving  into  to avoid importing types from a route module.\n\n### Small fixes applied\nI pushed a small follow-up commit to this PR branch: \n- Fix unit test relative import so it runs from repo root.\n- Avoid  inside the  loop.\n- Lazily initialize the invite banner interval to avoid redundant localStorage reads during render.\n- Update  doc comment to match current behavior.\n\n### Local verification\n- bun test v1.3.8-canary.1 (b64edcb4)\n\n✅ Approving — looks good to merge once CI is green.

slkzgm (approved):
## Review summary

This PR delivers the described UX changes:
- **Stories**: flat interleaved feed (no story header chrome) + new-market cards in the scored stream.
- **Latest**: posts + new markets merged chronologically (no longer pinned/batched).
- **Other tabs**: no market cards.

### What I checked
- Algorithms are extracted into pure utilities and are covered by unit tests.
- Client rendering paths are memoized (`useMemo`) where it matters.
- API routes return `marketId` + live share counts to support deep links and probability bars.

### Non-blocking concerns / follow-ups
- **DB perf**: joining `markets` via `lower(trim(question)) = lower(trim(text))` will not use a normal index. Given caching this is probably OK short-term, but we should migrate to a real FK (or add a functional index) as a follow-up.
- **Type boundaries**: consider moving `NewMarketEntry` into `@babylon/shared` to avoid importing types from a route module.

### Small fixes applied
I pushed a small follow-up commit to this PR branch: `2370e98fe`
- Fix unit test relative import so it runs from repo root.
- Avoid `indexOf()` inside the `flattenStories` loop.
- Lazily initialize the invite banner interval to avoid redundant localStorage reads during render.
- Update `useNewMarkets` doc comment to match current behavior.

### Local verification
- `bun test packages/testing/unit/narrative-feed-algorithms.test.ts`

✅ Approving — looks good to merge once CI is green.

PR #1184, opened 2026-03-04
fix(testing): resolve TypeScript errors in narrative-enrichment-cache tests

revlentless:
## Summary                                                
                                                                                                                                                                                                                                                                                                                                       
  - Fix 2 TypeScript errors in `packages/testing/unit/narrative-enrichment-cache.test.ts`. widen mock return type and fix parameter type incompatibility so `tsc -b .` passes cleanly


slkzgm (approved):
Summary
- Fixes the two TypeScript errors in `packages/testing/unit/narrative-enrichment-cache.test.ts` by widening the `mockGetCache` return type and avoiding the `opts` param type incompatibility in the `mockSetCache` implementation.

Notes / follow-ups
- `bun run typecheck` still fails on `packages/api/src/services/whitelist-service.ts` (TS2345: `SQL<unknown>` not assignable to `never`). This reproduces on `origin/staging` and is not introduced by this PR.
- I pushed an extra small commit (`c49f8065c`) to mirror production behavior where cache writes never block the response (i.e. `setCache(...).catch(...)`). This also fixes the existing failing test in this file.

Validation
- `bun run typecheck` (still failing only with the `whitelist-service.ts` errors above)
- `bun test unit/narrative-enrichment-cache.test.ts`

PR #1198, opened 2026-03-06
experiment: @babylon/sim - Standalone Tick Engine

revlentless:
## Why this exists

The production game tick (`executeGameTick()` in `@babylon/engine`) is a single ~900-line function that manually wires 17 subsystems with deadline-aware gating, result accumulation, and interval-gated operations. Every new feature requires editing that function and manually wiring services. There's no standardized way to add engine functionality, no per-subsystem error isolation, and no way to test subsystems independently.

`@babylon/sim` is a new package that provides a proper runtime for the simulation. Instead of one monolithic function, each piece of game logic becomes a **system** - a small, self-contained unit with a defined phase, dependencies, and lifecycle. The engine handles ordering, deadline gating, metrics, and error isolation so each system only worries about its own logic.

This is an **experiment branch**. The only change to existing code is adding an optional `skip` parameter to `executeGameTick()` (defaults to an empty set - zero behavior change for all current callers). Outside that, just a path alias in the root `tsconfig.json`.

### The current problem

```
executeGameTick() {
  // 900 lines of manually wired subsystems
  // One throws? Hope the try/catch is in the right place
  // Want to add something? Edit this file, wire it in, hope nothing breaks
  // Want to test just markets? You can't, it's all one function
  // Want to skip something when over budget? Manual if-checks everywhere
}
```

### What this enables

```ts
// systems/feed-generation.ts - that's it, drop a file
export default defineSystem({
  id: 'feed-generation',
  name: 'Feed Generation',
  phase: TickPhase.ContentMaintenance,
  dependencies: ['market-context'],

  async onTick(ctx) {
    // isolated, testable, auto-discovered
    return { metrics: { postsGenerated: 5 } };
  },
});
```

**Drop a file in `systems/`, export a `defineSystem()` call, and it runs.** No wiring, no editing a central file, no coordination.

- **Phase ordering + dependency resolution**: Systems declare what phase they belong to and what they depend on. The engine topologically sorts within each phase. No manual ordering.
- **Error isolation**: If a system throws, the engine logs it, fires a hook, adds a warning, and moves on. Other systems still run. The tick doesn't abort.
- **Deadline gating**: Each tick has a budget. When time's up, remaining systems are skipped (with warnings), except `Finalize`-phase systems and anything marked `skipDeadlineCheck`.
- **Per-system metrics**: Each system returns its own metrics. Numbers are summed across systems automatically. No manual accumulation.
- **Lifecycle hooks**: `engine:boot`, `tick:before`, `tick:after`, `system:error`, etc. Observable from outside or from within a system's `register()`.
- **Composables**: `useTick()`, `useDB()`, `useLLM()`, `useServices()` - call them anywhere during a tick, no prop drilling. Uses `AsyncLocalStorage` under the hood.
- **Type augmentation**: Systems declare their config keys, services, and shared data types via `declare module '@babylon/sim'`. Everything resolves at the call site without casting - same pattern as Nuxt/Nitro.

### Zero-risk migration path (hybrid mode)

The legacy bridge (`createLegacyGameTickSystem()`) wraps the entire existing `executeGameTick()` as a single system. Run it with `--legacy` and you get the exact same behavior as today, inside the new runtime. Then extract subsystems one at a time:

1. Start with `--legacy` running the full old tick
2. Extract one subsystem into its own system file
3. Add its ID to `migratedSubsystems` in `babylon.config.ts`
4. The bridge passes the skip set to `executeGameTick()` - old code skips it, new system handles it
5. Repeat until the bridge is empty

Both old and new code run in the same tick. No big-bang rewrite. No flag day.

```ts
// babylon.config.ts - as you migrate systems, add their IDs here
export default defineBabylonConfig({
  migratedSubsystems: ['widget-caches', 'trending-tags'],
});
```

## Architecture

Built on the [UnJS](https://unjs.io) stack:

| Library | Role |
|---------|------|
| [c12](https://github.com/unjs/c12) | Config loading from `babylon.config.ts`, dotenv from repo root |
| [citty](https://github.com/unjs/citty) | CLI framework (dev, build, tick, info, document) |
| [unimport](https://github.com/unjs/unimport) | Auto-discovery of system files via `scanDirExports` |
| [hookable](https://github.com/unjs/hookable) | Lifecycle hooks (engine extends `Hookable<RuntimeHooks>`) |
| [unctx](https://github.com/unjs/unctx) | Async-safe composables via `AsyncLocalStorage` |
| [chokidar](https://github.com/paulmillr/chokidar) | File watching for dev mode hot-reload |

### Tick phases

Systems run in phase order. Within a phase, dependencies are topologically sorted.

| Phase | Value | Purpose |
|-------|-------|---------|
| Bootstrap | 100 | Init, day number, LLM warmup |
| Questions | 200 | Question generation, oracle commits |
| Events | 300 | World events, narrative arcs |
| Markets | 400 | NPC trading, price updates |
| Rebalancing | 500 | Portfolio rebalancing |
| ContentMaintenance | 600 | Question top-up, arcs, timeframes |
| Social | 700 | Invites, relationships, groups |
| Finalize | 800 | Caches, trending, reputation, stats (always runs) |

### Type augmentation

Systems extend global interfaces via declaration merging - no codegen, no build step:

```ts
// In your system file
declare module '@babylon/sim' {
  interface BabylonConfig {
    myFeature: { apiKey: string; retries?: number };
  }
  interface BabylonServices {
    feedCache: Map<string, unknown>;
  }
  interface BabylonSharedData {
    feedReady: boolean;
  }
}
```

Now `ctx.config.myFeature`, `ctx.services.get('feedCache')`, and `ctx.shared.get('feedReady')` are all fully typed at every call site. When no system augments these interfaces, they fall back to untyped (`string` keys, `unknown` values) - no breaking change.

## What's in the box

```
packages/sim/
  babylon.config.ts          Config file
  systems/
    hello-world.ts           Example system with type augmentation
  core/
    engine.ts                BabylonEngine runtime (extends Hookable)
    types.ts                 All interfaces, TickPhase, conditional typed containers
    augments.ts              Empty interfaces for declaration merging
    system.ts                defineSystem() helper
    composables.ts           useEngine(), useTick(), useDB(), etc.
    config.ts                c12 config loading + .env from git repo root
    scanner.ts               unimport auto-discovery
    context.ts               EngineContext + TickContext factories
    metrics.ts               Per-tick metric accumulator
    service-container.ts     Simple DI container
    llm-orchestrator.ts      LLM wrapper
    errors.ts                FrameworkError, SystemNotFoundError, etc.
    bridge/
      legacy-game-tick.ts    Wraps executeGameTick() as a system
  cli/
    index.ts                 CLI entry (citty)
    shared.ts                Shared CLI utilities
    commands/                dev, build, tick, info, document
  docs/                      Hand-written guides
  Dockerfile                 Multi-stage Docker build (341MB image)
  docs/
    migration-plan.md        Full 24-system migration plan
  tests/                     208 tests across 13 files
```

## Test coverage

208 tests, 0 failures, covering:

- Engine lifecycle (boot, tick, shutdown, destroy order)
- Phase ordering + topological sort (including cycle detection)
- Deadline gating (skip, Finalize always runs, skipDeadlineCheck)
- Metric accumulation (increment, set, type mismatch warnings)
- Error isolation (system throws, tick continues)
- Composables (AsyncLocalStorage, useDB, useLLM, useServices, etc.)
- Hooks (all 7 lifecycle events, hookOnce, system:error)
- Service container (register, get, duplicate, missing)
- Config loading (defaults, c12, defineBabylonConfig)
- Scanner (missing dir, empty dir, dedup, non-system files)
- Intervals (tick-based, time-based, validation)
- Full integration (multi-system, dependencies, shared data flow)

All tests use mocks - no database, no API keys.

## CLI

```
babylon dev [--interval 60] [--legacy]     Dev mode with hot-reload
babylon tick [--loop] [--interval 60] [--legacy]  Single tick or loop
babylon build [--outDir .output] [--minify]       Bundle for production
babylon info                                       Print config + systems
babylon document [--outDir .docs]                  Generate markdown reference
```

## Standalone deployment

The build produces a self-contained 9MB bundle. The Dockerfile packages it into a 341MB image.

```bash
# Docker
bun run docker:build                              # build image
docker run -e SIM_ONCE=1 babylon-sim              # single tick (cron target)
docker run -e TICK_INTERVAL_MS=30000 babylon-sim   # loop mode

# Without Docker
babylon build && bun .output/server.js --once     # single tick
SIM_ONCE=1 bun .output/server.js                  # single tick (alt)
bun .output/server.js                             # loop mode
```

## Changes outside `packages/sim/`

- **`packages/engine/src/game-tick.ts`**: Added optional `skip` parameter to `executeGameTick()` - defaults to empty set, zero behavior change for existing callers. Logs skipped subsystem IDs when non-empty.
- **`tsconfig.json`**: Added `@babylon/sim` path alias.

## Not in scope

- This does **not** replace the current deployment - it runs alongside it
- This is an experiment to validate the architecture


slkzgm (approved):
Overall this is a strong foundational architecture change: a system-based tick runtime with phases, dependency ordering, deadline gating, per-system isolation, and AsyncLocalStorage-based composables. The core engine/scanner/hooks/metrics abstractions are clean, and the incremental extraction story (bridge wraps the legacy tick as one system) is reasonable.

I pushed a small follow-up commit to the PR branch (cf495649e) to align implementation with the documented config and to tighten DX:
- Wire systemPhases overrides when registering scanned systems (CLI) and in the generated build entry.
- Implement dev.watchConfig via c12 watcher; serialize tick/reload to avoid shutdown() racing an in-flight tick(); coalesce reload bursts.
- Tighten build entry registration to require { id, name, phase, onTick }.
- Clarify migratedSubsystems semantics (currently observability-only; executeGameTick() does not universally gate subsystems by this set yet).
- Prevent bun test from picking up compiled dist/tests in addition to tests by narrowing the pattern to tests/*.test.ts; update README accordingly.

Non-blocking follow-ups to consider:
- The PR description currently implies migratedSubsystems can be used as a true skip set for legacy subsystems; that is not true until executeGameTick() honors it. Either update the PR text/docs to avoid over-promising, or implement gating in the legacy tick when ready.
- Docker runtime stage installs deps via "bun add --no-save" which is non-deterministic; prefer baking required externals via lockfile/install stage, or avoid externalizing those deps.
- Potential CLI name collision: @babylon/sim defines a "babylon" bin, but @babylon/cli already owns "babylon". Consider renaming the sim binary to avoid ambiguity.

PR #1206, opened 2026-03-08
feat(feed): replace Stories with For You

SYMBaiEX:
## Summary
- replace the Stories feed with a dedicated For You recommendation surface
- keep Following, Trades, Latest, and Hot as separate filtered tabs
- move For You onto its own ranking pipeline with feed-event telemetry and diversity controls

## What changed
- add a standalone /api/feed/for-you pipeline instead of wrapping the old narrative endpoint
- rank mixed feed items using topic relevance, market relevance, social affinity, engagement velocity, conversation depth, urgency, freshness, and fatigue penalties
- add feed event capture for impressions, dwell, opens, likes, shares, and trade-after-view
- wire the For You client to emit ranking signals from feed interactions
- add FeedEvent schema and migration metadata for the new telemetry table
- keep the legacy narrative utilities in place for now as follow-up cleanup

## Verification
- bun test apps/web/src/app/api/feed/events/route.test.ts apps/web/src/app/api/feed/for-you/route.test.ts apps/web/src/app/api/feed/for-you/scoring.test.ts packages/testing/unit/narrative-feed-algorithms.test.ts

## Notes
- includes a small engine typecheck fix restoring the missing eq import in the daily topic service
- repo-wide typecheck cleanup outside this slice is intentionally not bundled here


slkzgm (changes requested):
Requesting changes for one blocking performance regression and one maintainability concern.

1. The authenticated `/api/feed/for-you` path currently does 6 per-user DB reads on every refresh (`follows`, `userActorFollows`, `reactions`, `shares`, `positions`, `feedEvents`) with no short-lived enrichment cache. The previous narrative feed kept the user-specific enrichment behind a 30s cache for exactly this reason. Since For You is now the default tab and refreshes on SSE plus manual pulls, this is a meaningful load regression on the hottest feed path. Please reintroduce a short TTL per-user enrichment cache (or an equivalent reuse of the existing cache helper) before merge.

2. `useForYouFeed` is effectively a fork of `useNarrativeFeed` with the same control flow, race handling, interval/SSE logic, and error semantics. That duplication is already enough to create maintenance drift. Please extract the shared fetching state machine into a reusable hook/helper instead of carrying two nearly identical implementations.

revlentless:
Rebased onto staging and pushed fixes for both blocking items:

1. **Per-user enrichment cache** — All 6 per-user DB queries in the For You pipeline are now wrapped in `getCacheOrFetch` with a 30s TTL (`feed:for-you:enrichment:{userId}`), matching the narrative feed's caching behavior.

2. **Shared feed hook** — Extracted `useFeed` from `useForYouFeed`/`useNarrativeFeed`, removing ~400 lines of duplicated state machine code. Both hooks are now thin config wrappers.

Also fixed: retry off-by-one in `useFeedEventTracker` (was allowing 4 attempts instead of 3) and a `LikeButton` bug where selecting a different reaction on an already-liked post would toggle it off.

slkzgm (changes requested):
Requesting changes for two remaining implementation gaps in the current head.

1. `loadBaseCandidates()` still materializes the synthetic `__general__` story and then also materializes standalone `post:${post.id}` items from the exact same `generalPosts` bucket. Because `buildForYouFeed()` later collapses each story to a single lead post, the top general post can now appear twice in the returned feed: once as the lead item of `__general__` and again as its standalone card. That is user-visible duplication and it also biases ranking because the same content competes twice. The pipeline should choose one representation for non-question posts, not both.

2. New-market cards are only partially wired into the new telemetry/interaction path. `ForYouFeedList` forwards `onOpenMarket`/`onTradeComplete`, but `NewMarketCard` renders its own `InteractionBar` with hard-coded zero counts and no `onLikeChange` / `onShareChange` callbacks. As a result, likes/shares/comments from these cards do not flow into `/api/feed/events`, and the card UI is not hydrated from the anchor post's real interaction state. That leaves one of the new feed item types outside the feedback loop this PR is introducing.

The earlier blockers around per-user enrichment caching and the shared feed hook look addressed in the current branch. Happy to re-review once these two are fixed.

revlentless:
@slkzgm 

Addressed both items from your second review:

1. **Duplicate general posts** — `__general__` story is now skipped in the pipeline loop. General posts only surface as standalone `post:{id}` cards, no more double-ranking.

2. **Market card telemetry** — `NewMarketCard` now passes `onLikeChange`/`onShareChange` through to its `InteractionBar`, and the interaction counts are hydrated from the anchor post instead of hardcoded zeros. Likes/shares from market cards now flow into `/api/feed/events` like everything else.

Also rebased on latest staging. Ready for re-review.

slkzgm (commented):
Re-reviewing the latest fix: the duplicate `__general__` ranking issue is resolved, and the new-market like/share telemetry wiring is now present.

There is still one blocking gap left in the interaction-state hydration path for market cards:

- `NewMarketCard` now tries to hydrate `InteractionBar` from `story.posts.find((p) => p.id === story.anchorPostId)`, but `buildForYouFeed()` still explicitly returns `posts: story.isNewMarket ? [] : leadPosts` for market stories. In other words, the lookup can never succeed for `isNewMarket` items, so the card still falls back to zero counts / false flags in practice.
- Concretely, the telemetry half of the fix landed, but the UI-state half is still ineffective until the pipeline carries anchor post interaction data through for market stories (or exposes those fields separately).

Once that is fixed, this looks ready from my side.

slkzgm (changes requested):
The earlier blockers around per-user enrichment caching, the shared feed hook, duplicate `__general__` stories, and market-card telemetry wiring look addressed on the current head.

There is still one blocking correctness issue in the latest fix:

- `loadBaseCandidates()` now returns `anchorPostById: Map<string, NarrativePost>`, and `buildForYouFeed()` puts that object behind `getCacheOrFetch('feed:for-you:v2:base', ...)`. The cache layer serializes values with `JSON.stringify` and reads them back with `JSON.parse`, so `Map` is not preserved across Redis hits. On a real cache hit, `baseResult.anchorPostById` will not be a `Map`, which means the new-market anchor hydration path is broken in production and can fail when `.entries()` is used.

Please keep the cached payload JSON-serializable (for example store anchor posts as an array/object and reconstruct a `Map` after reading, or keep the anchor-post hydration data outside the shared cache). Once that is fixed, I’m happy to take one more pass.

SYMBaiEX:
### Code review

Found 1 issue:

1. `EventAggregates` Maps are not JSON-serializable and will be corrupted on enrichment cache hit

The user enrichment data (including the 6th element — `EventAggregates` returned by `loadFeedEventAggregates`) is passed through `getCacheOrFetch`, which uses `JSON.stringify`/`JSON.parse` for Redis-backed caching. `Map` instances serialize as `{}`, so on any cache hit within the 30-second TTL, `eventAggregates.authorAffinity` (and all other maps) become plain objects. Every subsequent call to `getAffinityScore` or `getFatiguePenalty` then throws `TypeError: map.get is not a function`, crashing the For You feed for all authenticated users whose enrichment is served from cache.

The author was already aware of this category of bug and fixed it for `anchorPostById` in the `baseResult` cache (comment at line 1124: "Maps are not JSON-serializable, so reconstruct the Map here after the cache read"), but the same fix was not applied to the `EventAggregates` Maps in the enrichment tuple.

https://github.com/BabylonSocial/babylon/blob/b0521524deffdedd3de7f833be5a395096db7784/apps/web/src/app/api/feed/for-you/pipeline.ts#L1017-L1093

The enrichment cache fetch (lines 1027–1086) returns a 6-element tuple; the 6th element at index 5 is the `EventAggregates` object. After the cache read, only `anchorPostById` (in the separate `baseResult` cache) is reconstructed into a `Map`. The `eventAggregates` Maps are used directly at lines 1189–1256 without reconstruction.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1215, opened 2026-03-10
fix(chat): add NPC guard for existing DM chats

0xbbjoker:
## Summary
- Adds NPC actor check when sending messages to **existing** DM chats. The auto-create path already blocked DMs to NPCs, but the existing-chat path only checked block status — allowing bypass via a pre-existing `dm-{userId}-{npcId}` chat record.
- Queries `users.isActor` in parallel with block checks to avoid extra round-trips.

## Test plan
- [ ] Verify `POST /api/chats/{dm-userId-npcId}/message` returns error when other participant is NPC
- [ ] Verify normal DM messaging still works
- [ ] Verify `POST /api/chats/dm` still rejects NPC targets (unchanged)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -115,6 +115,7 @@ scripts/cleanup-excess-markets.ts
 
 # Continuous Claude cache (local only)
 .claude/cache/
+.claude/worktrees/
 .claude/tsc-cache/
 .claudekit/
 .env.hook-test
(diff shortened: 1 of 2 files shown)


slkzgm (approved):
Core fix is correct: the existing-DM path now applies the same NPC guard as the auto-create path, and the added lookup is efficiently batched with the block checks.

I pushed a small cleanup commit to remove the accidental Claude worktree gitlinks and added the local worktree path to .gitignore to prevent recurrence.

I did not find any remaining blocking issues. The only follow-up I would still recommend is adding an automated regression test for this bypass in a separate change.

PR #1217, opened 2026-03-10
fix(mcp): correct double /api prefix in get_post URL

0xbbjoker:
## Summary
- `getAPIBaseUrl()` already returns `baseUrl/api`, so `executeGetPost` was constructing `baseUrl/api/api/posts/:id` — a double `/api` prefix causing 404 for valid post IDs returned by `query_feed`
- Removed the redundant `/api` segment so the URL is `baseUrl/api/posts/:id`

## Test plan
- [ ] Call `babylon_get_post` with a post ID from `babylon_get_feed` and verify it returns post data
- [ ] Verify no regressions in other post operations (create, delete, like)

### DIFF
--- a/packages/mcp/src/handlers/tool-handlers.ts
+++ b/packages/mcp/src/handlers/tool-handlers.ts
@@ -1447,7 +1447,7 @@ export async function executeGetPost(
 ): Promise<GetPostResult> {
   const apiBaseUrl = getAPIBaseUrl();
   return safeFetchRequired<GetPostResult>(
-    new URL(`${apiBaseUrl}/api/posts/${args.postId}`)
+    new URL(`${apiBaseUrl}/posts/${args.postId}`)
   );
 }
 


slkzgm (approved):
The fix is correct and matches the PR description:  already includes , so removing the extra segment from  fixes the broken  URL without adding extra complexity.

Non-blocking follow-up: while reviewing I confirmed there are pre-existing manual URL constructions in the same file that still append  after  (not introduced by this PR), notably the portfolio and leaderboard fetch paths. They should be cleaned up separately to avoid the same class of bug elsewhere.

slkzgm (approved):
The fix is correct and matches the PR description: `getAPIBaseUrl()` already includes `/api`, so removing the extra segment from `executeGetPost()` fixes the broken `babylon_get_post` URL without adding extra complexity.

Non-blocking follow-up: while reviewing I confirmed there are pre-existing manual URL constructions in the same file that still append `/api` after `getAPIBaseUrl()` (not introduced by this PR), notably the portfolio and leaderboard fetch paths. They should be cleaned up separately to avoid the same class of bug elsewhere.

PR #1218, opened 2026-03-10
fix: prevent 500 from garbage/malformed JWT tokens

0xbbjoker:
## Summary
- Wrap `verifyAgentSession()` in try-catch so Redis/store errors fall through to Privy auth instead of leaking as 500
- Wrap `getPrivyClient()` in try-catch that throws `AuthenticationError` (401) instead of plain `Error` (500) when Privy credentials are missing
- Garbage JWTs now consistently return 401 instead of 500

## QA Bug
**C2**: Garbage/malformed JWT returns 500 instead of 401

## Test plan
- [ ] Send request with garbage Bearer token → should get 401
- [ ] Send request with expired JWT → should get 401 with expiration message
- [ ] Send request with valid JWT → should authenticate normally
- [ ] Verify no regressions in agent session auth flow


slkzgm (changes requested):
I found two blocking issues in the current implementation:

1. `optionalAuth()` and `optionalAuthFromHeaders()` still call `verifyAgentSession()` without the new guard. That means the same store failure you fixed in `authenticate()` can still leak out of the optional-auth paths, and those paths are used by multiple API routes. The result is an incomplete fix and inconsistent behavior across the auth surface.

2. Wrapping `getPrivyClient()` and converting missing Privy credentials into `AuthenticationError` changes a server-side configuration failure into a `401`. That does not represent the failure correctly and it hides an operational problem behind an auth response. If we want to avoid a raw `Error`, this should be mapped to an existing 5xx boundary error (for example `ServiceUnavailableError`) rather than an authentication failure.

Please address those two points and add focused coverage for the new failure paths when you update the PR.

0xbbjoker:
Claude says: All three points from the review have been addressed:

1. **`optionalAuth()` and `optionalAuthFromHeaders()`** — Both now wrap `verifyAgentSession()` in try-catch (commits `fdc4e61` and `d14c844`), so store failures return `null` instead of leaking as 500.

2. **`getPrivyClient()` failure mapping** — Changed from `AuthenticationError` (401) to `ServiceUnavailableError` (503) in `authenticate()`. This correctly represents a server-side configuration failure rather than masking it as an auth error (commit `666f164`).

3. **Test coverage** — Added tests for the `verifyAgentSession` throw fallback path and the `getPrivyClient` failure path, including a `_resetPrivyClientForTesting()` helper to properly reset the lazy singleton between tests.

Claude

revlentless:
LGTM. The core fix is solid and all three auth paths (authenticate, optionalAuth, optionalAuthFromHeaders) now handle verifyAgentSession and getPrivyClient failures consistently. Added tests for the optionalAuth and optionalAuthFromHeaders failure paths to round out coverage. Good to merge.

PR #1219, opened 2026-03-10
fix(mcp): ensure create_post returns mediaUrl from args directly

0xbbjoker:
## Summary
- `executeCreatePost` was returning `post.imageUrl ?? null` from the DB result, which could be null if the ORM doesn't populate the column in the `.returning()` result
- Now uses the validated `args.mediaUrl` value directly in the response, ensuring the client always sees the media URL they submitted

## Test plan
- [ ] Call `babylon_create_post` with `mediaUrl` parameter and verify response includes `mediaUrl`
- [ ] Call `babylon_create_post` without `mediaUrl` and verify `mediaUrl` is null in response

### DIFF
--- a/packages/mcp/src/handlers/tool-handlers.ts
+++ b/packages/mcp/src/handlers/tool-handlers.ts
@@ -1461,13 +1461,14 @@ export async function executeCreatePost(
   await checkMcpRateLimit(agent.userId, 'post');
 
   const postId = await generateSnowflakeId();
+  const mediaUrl = args.mediaUrl ?? null;
   const post = await db.post.create({
     data: {
       id: postId,
       content: args.content.trim(),
       authorId: agent.userId,
       type: args.type || 'post',
-      imageUrl: args.mediaUrl ?? null,
+      imageUrl: mediaUrl,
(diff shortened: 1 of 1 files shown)


slkzgm (approved):
Approved. The code change is small, low-risk, and it does make the MCP response consistently return the validated media URL that was just submitted.\n\nOne note from the local code review: the current  path already goes through , which uses Drizzle  for inserts, so the PR description's rationale about the ORM not returning  does not appear to match the current implementation. The change itself is still fine, but the problem statement is likely overstated or based on an older code path.\n\nI also did not find an existing lightweight handler test covering . A follow-up targeted test for the  /  response contract would help lock this behavior in.

slkzgm:
Follow-up clarification to my approval review: db.post.create in this repo goes through TableRepository#create, and TableRepository#create already uses Drizzle returning() on insert. So the code change is fine, but the PR description rationale about the ORM not returning imageUrl does not match the current implementation. I also did not find a lightweight existing test covering executeCreatePost; a focused test for the mediaUrl/null contract would be useful.

PR #1220, opened 2026-03-10
fix: omit undefined profile update fields

0xbbjoker:
## Summary
- filter undefined values out of the profile update payload before calling Drizzle 
- prevent  from sending undefined fields into the update query
- keep partial profile updates working without triggering the staging 500

## Validation
- bun test packages/testing/unit/web/api-routes-with-error-handling.test.ts --preload ./packages/testing/unit/preload.ts
- bun run typecheck
- bun run lint
- bun run build *(fails in existing  Turbopack workspace-root detection, unrelated to this change)*


slkzgm (dismissed):
Reviewed the PR end-to-end, including the description, bot conversations, and the code diff. The core fix is correct: filtering undefined-derived fields before Drizzle update() is the right minimal change for the staging 500. I pushed a small follow-up commit (7586fb1a2) to address two review issues discovered during validation: the new unit test was using a Bun mock pattern that does not resolve reliably for workspace barrel exports, and the no-op path should not emit profile-update audit/analytics events for an empty payload. With that follow-up in place, the targeted unit coverage passes and the implementation looks consistent with the existing route patterns.

PR #1221, opened 2026-03-10
fix: harden positions proceeds preview

0xbbjoker:
## Summary
- harden the positions API against invalid sell-preview math for prediction positions
- fall back to cost basis instead of throwing when proceeds cannot be calculated
- add a unit test covering non-finite sell proceeds

## Validation
- bun test packages/testing/unit/prediction-amm.test.ts
- pre-push checks passed ( and downstream checks run by the repo hook)

## Notes
- warning-only build messages were emitted during repo hooks, but checks completed successfully
- untracked local directories like , , and  were not included in this branch


revlentless:
LGTM. Clean extraction of the calculation into its own helper, good fallback to cost basis instead of letting the 500 leak, and the structured logging gives visibility when it triggers. The minor currentUnitPrice change in the early-return path (avgPrice vs costBasis/shares) only affects the edge case of an empty pool, and the difference is ~0.1% at current fee rates, so not a concern. Good to merge.

PR #1231, opened 2026-03-12
Add permissions for PRR job in workflow

odilitime:


revlentless:
LGTM. Permissions are correctly scoped at the job level, fork PRs are excluded, and it uses the auto-scoped GITHUB_TOKEN. One thing to confirm: does the PRR reusable workflow actually need `contents: write`, or would `contents: read` work? The claude-code-review workflow only uses `contents: read` for similar review functionality. If PRR only reads code and posts reviews, read is safer. If it needs to push suggestion commits, write is correct. Either way, the change itself is clean and good to merge.

PR #1233, opened 2026-03-12
feat: development mode theme toggle

revlentless:
## Summary
- Adds a floating theme toggle button (bottom-right corner) that only renders in development (`NODE_ENV === 'development'`)
- Toggles between light and dark mode with a single click, replacing the need to navigate to the settings page during development
- Component is mounted inside `ThemeProvider` in both Providers branches (with and without Privy)

## Test plan
- [x] Run the app locally (`NODE_ENV=development`) and verify the floating button appears in the bottom-right corner
- [x] Click the button and confirm it toggles between light and dark themes
- [x] Verify the button does not appear in production builds (`next build && next start`)
- [x] Confirm existing theme settings page still works independently

### DIFF
--- a/apps/web/src/components/providers/Providers.tsx
+++ b/apps/web/src/components/providers/Providers.tsx
@@ -7,6 +7,7 @@ import { useTheme } from 'next-themes';
 import { Fragment, Suspense, useEffect, useRef, useState } from 'react';
 import { PostHogErrorBoundary } from '@/components/analytics/PostHogErrorBoundary';
 import { PostHogIdentifier } from '@/components/analytics/PostHogIdentifier';
+import { DevThemeToggle } from '@/components/shared/DevThemeToggle';
 import { ThemeProvider } from '@/components/shared/ThemeProvider';
 import { FontSizeProvider } from '@/contexts/FontSizeContext';
 import { WidgetRefreshProvider } from '@/contexts/WidgetRefreshContext';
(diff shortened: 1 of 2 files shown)


slkzgm (changes requested):
The dev-only theme toggle is a useful DX improvement, but I do not think this is ready to merge in its current placement.

Blocking:
- `DevThemeToggle` is mounted at `fixed right-4 bottom-4` with `z-[9999]`, which directly overlaps `GameOnboardingProgress` (`apps/web/src/components/onboarding/GameOnboardingTooltip.tsx`) when onboarding is active. In dev, that makes the onboarding progress UI partially obscured.
- The same placement also sits on top of the mobile `BottomNav` (`apps/web/src/components/shared/BottomNav.tsx`). Because the toggle has a much higher z-index, it can intercept taps over the nav area on small screens.

Please move the control to a location that does not conflict with existing fixed UI, or offset it using the existing layout primitives (for example `--bottom-nav-height` on mobile) rather than introducing another ad hoc overlay.

Non-blocking:
- Please add an `aria-label` to the button for consistency with the rest of the codebase and basic accessibility.

revlentless:
@slkzgm good call. I switched it to a shortcut `alt+shift+T`

slkzgm (approved):
The previous blocking issue is resolved: switching from a floating fixed control to a dev-only keyboard shortcut removes the overlap with onboarding and mobile navigation.

I do not see any remaining blocking code issues in the updated implementation.

Non-blocking follow-up:
- Please update the PR description and test plan to reflect the current behavior (`Alt+Shift+T` keyboard shortcut) instead of the original floating button wording.

PR #1234, opened 2026-03-12
feat: add achievements & challenges system with Figma UI, SSE notifications, and daily rewards

revlentless:
## Summary

- Adds a full V1 achievements & challenges engine: 15 permanent achievements (8 Bronze, 5 Silver, 2 Gold) and 40 rotating challenges (20 daily pool → 3/day, 20 weekly pool → 2/week) with query-based progress tracking, completion bonuses (+40 daily, +100 weekly), and points integration
- New Figma-matched rewards UI with 3-tab layout (Overview, Achievements, Challenges), daily streak card with claim flow, live countdown timers, tier filtering, progress bars, and SSE-powered real-time UI refresh + toast notifications
- Fire-and-forget `checkProgress()` wired into 16 route handlers covering trades, posts, comments, agents, follows, groups, page visits, daily login, and market resolution

## Type

- [x] Feature

## Context / Links

- Implements the full achievements & challenges system from design specs
- Achievement UI matches Figma designs in `docs/achievement-ui/`

## Scope (keep it focused)

- [x] This PR is focused on a single change/theme (not a catch-all)
- [x] Drive-by refactors are excluded or split into a separate PR
- [x] Non-goals / follow-ups are listed below (with links)

**Areas touched**
- [x] Web UI (`apps/web`)
- [x] Web API routes / SSE / A2A (`apps/web`)
- [x] API infra (`packages/api`)
- [x] DB (`packages/db`)
- [x] Shared types/utils (`packages/shared`)
- [x] Tests (`packages/testing`)
- [x] Domain / game engine (`packages/engine`, `packages/core/*`)

### Backend (packages/api, packages/db, packages/shared, packages/engine)
- **Achievement service** (`packages/api/src/services/achievement-service.ts`): ~1600-line engine with `checkProgress()`, `getUserAchievements()`, `getUserChallenges()`, query-based progress resolvers, deterministic SHA-256 challenge rotation, and SSE broadcasts
- **DB schema**: New tables `AchievementDefinition`, `ChallengeDefinition`, `UserAchievement`, `UserChallengeProgress` with indexes and unique constraints (migration 0045)
- **Shared constants**: 15 achievement definitions + 40 challenge definitions in `packages/shared/src/constants/achievements.ts`, plus `POINTS.CHALLENGE_DAILY_ALL_BONUS` and `POINTS.CHALLENGE_WEEKLY_ALL_BONUS`
- **API routes**: `GET /api/achievements`, `GET /api/challenges` with authenticated access
[...]

PR #1244, opened 2026-03-13
feat: relaunch achievements & challenges system

slkzgm:
## Summary

Replacement PR for #1234 after the original merge was reverted from `production` because Vercel rejected the deployment.

This branch reapplies the full achievements + challenges feature set, including the review fixes that landed before the revert:
- verified heartbeat auth for achievement progress
- correct profile-owner achievements endpoint for `RecentAchievements`
- deduped challenge completion bonus award path

## Why This PR Exists

The original merge of #1234 was reverted from `production` after the post-merge deployment failed with:
- `Error: A Serverless Function has exceeded the unzipped maximum size of 250 MB.`

Relevant links:
- Original PR: #1234
- Production revert commit: `ebc841e6d3b2e9085f86944db6b926788e961fcf`
- Failed deployment: https://vercel.com/eliza-os/babylon/7LU38neAx551g91KqjyCktZwVVRd
- Vercel docs: https://vercel.link/serverless-function-size
- Vercel KB: https://vercel.com/kb/guide/troubleshooting-function-250mb-limit

## Current Status

This PR is intentionally opened as a **draft**.

The feature is not ready to re-merge until we identify the exact serverless function bundle that crossed Vercel's 250 MB unzipped limit and reduce it.

## Current Investigation Notes

Best current hypothesis from repo inspection:
- `apps/web` transpiles internal workspace packages directly via `transpilePackages` in `apps/web/next.config.ts`
- `@babylon/api` is exported through a very large barrel entrypoint in `packages/api/src/index.ts`
- this feature wires `checkProgress()` into many route handlers, which may cause one or more serverless functions to trace/bundle a significantly larger slice of `@babylon/api` + the achievement engine than before

This is not yet confirmed root cause, but it is the highest-signal area to inspect first.

## Suggested Next Steps

- identify the exact serverless function(s) that exceed 250 MB unzipped
- inspect traced files / bundle composition for that function
- reduce bundle size, likely by narrowing imports and/or avoiding large barrel entry points on hot route handlers
- rerun deployment validation before marking this PR ready

## Notes

Please use this PR for the relaunch discussion rather than the already merged/reverted #1234 thread.


revlentless:
@slkzgm i will take a look into whats causing that soon.
i noticed some things off in the ui, not everything is dark and light mode themed, so lets keep this as draft for that reason as well

slkzgm:
Review note (would be **Request changes** if GitHub allowed it on the author's own PR):

1. `packages/api/src/services/achievement-service.ts:1191-1246` can double-award challenge completion points under concurrency. The row is read first, then updated/inserted, and the reward path runs whenever `completed && !existing[0]?.completed`. On the insert path, `onConflictDoNothing()` is ignored, so a conflicting insert still falls through to `awardPoints()` / notification / SSE. On the update path, two requests that both observed `completed = 0` can both update the same row and both award points. This needs to be gated on an actual state transition in the database (transactional upsert, conditional update with `returning`, or equivalent).

2. The rewards UI is duplicated in this PR. `apps/web/src/app/rewards/page.tsx:6-9` only mounts the `rewards/v2` components, but the PR also adds a full parallel `apps/web/src/components/rewards/rewards/*` tree plus tests that still target that unused tree (`apps/web/src/components/rewards/rewards/__tests__/rewards-tabs.test.ts:1-7`). That leaves the shipped UI effectively untested while adding a large dead maintenance surface. I would either remove the unused implementation or move the tests to the live `v2` components before merging.

Non-blocking follow-up: `getUserAchievements()` and `getUserChallenges()` still resolve progress sequentially (`packages/api/src/services/achievement-service.ts:1375-1501`). The page now hits both endpoints on load and on SSE refreshes, so batching/caching those resolvers would materially reduce query load.

slkzgm:
Follow-up review after `fb42f342`: the two blockers from my previous pass look addressed.

- `checkChallenges()` now gates challenge rewards on an actual DB state transition by using conditional `update(...).returning()` / `insert(...).returning()` and only awarding when this request won the `completed = 0 -> 1` transition. That resolves the double-award race I flagged earlier.
- The unused `rewards/rewards/*` tree has been removed and the helper tests now point at `rewards/v2`, so the live UI is the one being exercised.

I do not have a new blocking code issue from this pass. Remaining note stays the same as a non-blocking follow-up: the achievements/challenges read endpoints still do sequential resolver queries, so there is still room to reduce query load.

From a code-review perspective this pass is in good shape; I would just wait for CI and Vercel to finish before considering it ready.

PR #1245, opened 2026-03-13
fix(coordinator): respect caller temperature and strengthen dispatch rules

revlentless:
## Summary
- **TEXT_SMALL temperature bug**: The Groq plugin's `TEXT_SMALL` handler hardcoded `temperature=0.7`, silently ignoring the coordinator's `temperature: 0.4`. This caused flaky instruction-following — e.g. the coordinator refusing to dispatch for "tell my agent to buy TSLAI for $100" and instead telling the user to @mention their agent.
- **Decision template improvements**: Added explicit "my agent" auto-resolve rule (look up team members, never tell user to @mention), added "how are my agents doing" as a dispatch trigger, and updated examples to show agent resolution from the team list.

## Changes
- `packages/agents/src/plugins/groq.ts` — `TEXT_SMALL` handler now accepts `temperature`, `frequencyPenalty`, `presencePenalty`, `maxTokens` from caller params (with same defaults as before)
- `packages/agents/src/plugins/plugin-user-core/src/coordinator-decision-template.ts` — Strengthened dispatch rules with auto-resolve and anti-@mention guidance

## Test plan
- [ ] Send "tell my agent to buy TSLAI for $100" in team chat → should dispatch to the user's agent, not tell them to @mention
- [ ] Send "how are my agents doing" → should dispatch to agent for status update
- [ ] Send "buy TSLAI" → should auto-resolve to user's single agent and dispatch
- [ ] Verify coordinator still handles greetings, data queries, and multi-agent scenarios correctly

### DIFF
--- a/apps/web/src/app/api/agents/team-chat/coordinator/route.test.ts
+++ b/apps/web/src/app/api/agents/team-chat/coordinator/route.test.ts
@@ -451,7 +451,9 @@ describe('Coordinator POST', () => {
 
       mockProcessActions.mockImplementation(
         async (_m: unknown, _a: unknown, _s: unknown, cb: Function) => {
-          await cb([{ content: { success: true, text: 'Dispatch completed' } }]);
+          await cb([
+            { content: { success: true, text: 'Dispatch completed' } },
+          ]);
         }
       );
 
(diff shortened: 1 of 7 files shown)


slkzgm (approved):
Reviewed the PR description, diff, and discussion. The functional fix is correct: TEXT_SMALL now respects caller-provided generation params instead of silently forcing 0.7, and the coordinator template changes align with the intended dispatch UX improvement. I also pushed a follow-up commit (0cd24a16e) adding regression coverage for the TEXT_SMALL parameter pass-through and default preservation so this does not silently regress again. No blocking issues from my side.

PR #1246, opened 2026-03-13
fix(coordinator): robust agent resolution and stronger dispatch rules

revlentless:
## Summary
- **Fuzzy agent name resolution**: The LLM often passes usernames, display names, or partial names instead of UUIDs. Added 4-tier matching: exact → normalized (strips spaces/punctuation, so "larry david" matches "larrydavid") → partial (substring match) → single-agent fallback (if user has only 1 agent, use it). Previously only exact match existed, causing "agent not found" failures.
- **Stronger dispatch prompt**: Rewrote the decision guide to hammer home auto-resolve behavior. Added explicit "NEVER tell the user to @mention" anti-pattern, "how are my agents doing" as a dispatch trigger, and emphasized copying the `[id: ...]` from Team Members.
- **Fix silent failure masking**: 3 locations had `?? true` that defaulted missing action results to success. Changed to `?? false` so failures surface properly.
- **Summary reliability**: Lowered summary temperature from 0.7 to 0.4, added anti-@mention rule to summary template, fixed fallback message that told users to @mention.

## Changes
- `packages/agents/src/services/AgentChatService.ts` — 4-tier fuzzy agent resolution with logging
- `packages/agents/src/plugins/plugin-user-core/src/coordinator-decision-template.ts` — Rewritten dispatch rules
- `packages/agents/src/plugins/plugin-user-core/src/actions/dispatch-to-agent.ts` — Better error messages with retry hint
- `apps/web/src/app/api/agents/team-chat/coordinator/route.ts` — ?? false fix, summary temp, anti-@mention
- `apps/web/src/app/api/agents/[agentId]/chat/route.ts` — ?? false fix

## Test plan
- [ ] "tell my agent to buy TSLAI for $100" with 1 agent → dispatches automatically
- [ ] "tell larrydavid to buy TSLAI" where username is "larrydavid" → resolves and dispatches
- [ ] "tell larry david to check positions" where username is "larrydavid" → fuzzy match resolves
- [ ] "how are my agents doing" → dispatches for status update
- [ ] "tell unknownbot to do something" → clear error with available agent list
- [ ] All existing tests pass (fast-path: 110, route: 15, dispatch: 30, groq: 2)

### DIFF
--- a/apps/web/src/app/api/agents/[agentId]/chat/route.ts
+++ b/apps/web/src/app/api/agents/[agentId]/chat/route.ts
@@ -582,7 +582,7 @@ export const POST = withErrorHandling(
               const firstResult = resultsArray[0];
               if (firstResult) {
                 actionResult = {
-                  success: firstResult.content?.success ?? true,
+                  success: firstResult.content?.success ?? false,
                   text:
                     typeof firstResult.content?.text === 'string'
                       ? firstResult.content.text
--- a/apps/web/src/app/api/agents/team-chat/coordinator/route.ts
+++ b/apps/web/src/app/api/agents/team-chat/coordinator/route.ts
@@ -167,6 +167,7 @@ No actions were taken this turn.
 4. Do NOT make up information — only reference data from the Actions You Completed section.
 5. If you dispatched to an agent, include a brief quote or summary of what the agent actually did or said. Use plain @username for mentions.
 6. Keep your response concise and factual.
+7. NEVER tell the user to @mention or tag their agents. The coordinator handles all dispatch routing.
 
 Output ONLY this XML:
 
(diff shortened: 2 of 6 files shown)


slkzgm (approved):
Reviewed the implementation end-to-end against the PR description, conversations, and diff. The main UX issue is addressed: agent-name resolution is more robust, dispatch guidance is materially clearer, and the silent success masking fix is directionally correct.

I pushed one follow-up commit to the PR branch: 0a98e3237 (). It tightens the new fallback behavior so we do not silently dispatch to the wrong agent when the user names an unknown agent, and it avoids arbitrary first-match routing on ambiguous partial matches. I also added focused unit coverage for normalized matching, generic single-agent fallback, unknown-name rejection, and ambiguous partial rejection.

Targeted verification run locally:
- bun test v1.3.9 (cf6cdbbb)
- bun test v1.3.9 (cf6cdbbb)
- 

One note on CI: the earlier  failure appears to be a repo-wide env audit issue ( reported as used-but-undeclared), not something introduced by this PR.

PR #1248, opened 2026-03-13
fix: skip elizaOS runtime migrations, manage tables via Drizzle

0xbbjoker:
## Problem

elizaOS `plugin-sql` runtime migrator (`RuntimeMigrator`) runs on **every** `runtime.initialize()` call. It:

1. Creates a `migrations` schema with `_migrations`, `_journal`, `_snapshots` tables
2. Introspects the entire DB, diffs against plugin schemas, and applies DDL changes
3. Takes **~2 minutes per agent** on cold start

This is called **3 times** in `AgentRuntimeManager` (NPC agents, user-controlled agents, coordinator), and once per unique agent runtime creation. In a serverless environment this is a non-starter — it was never designed for it.

Meanwhile, Babylon already manages its own schema via Drizzle Kit (`bun run db:generate && bun run db:migrate`) and stubs out the elizaOS DB adapter entirely (`createAdapterStubs`).

## Solution

**1. Skip runtime migrations** — pass `{ skipMigrations: true }` to all 3 `runtime.initialize()` call sites (same approach used by [eliza-cloud-v2](https://github.com/elizaOS/eliza-cloud-v2)).

**2. Own the elizaOS tables via Drizzle** — re-export `@elizaos/plugin-sql` schema tables in `packages/db/src/schema/eliza.ts` so they are included in Babylon's Drizzle Kit migrations. This means `bun run db:generate` will pick up any elizaOS table changes and `bun run db:migrate` will apply them in a controlled way.

## Changes

- `packages/agents/src/runtime/AgentRuntimeManager.ts` — `{ skipMigrations: true }` on all 3 `runtime.initialize()` calls
- `packages/db/src/schema/eliza.ts` — new file re-exporting plugin-sql schema tables (prefixed with `eliza*` to avoid name collisions)
- `packages/db/src/schema/index.ts` — exports the new eliza schema
- `packages/db/package.json` — adds `@elizaos/plugin-sql` as explicit dependency

## Migration path

After merging, run `bun run db:generate` to capture the elizaOS tables in a Drizzle migration, then `bun run db:migrate` to apply. Existing databases that already have these tables (created by the runtime migrator) will be a no-op.

### DIFF
--- a/packages/agents/src/runtime/AgentRuntimeManager.ts
+++ b/packages/agents/src/runtime/AgentRuntimeManager.ts
@@ -669,7 +669,11 @@ export class AgentRuntimeManager {
 
     // Initialize runtime to signal services that runtime is ready
     // This prevents 30s timeout errors in services waiting for runtime initialization
-    await runtime.initialize();
+    // Skip ElizaOS plugin-sql migrations — Babylon manages its own schema via Drizzle.
+    // ElizaOS tables are included in packages/db/src/schema/eliza.ts and migrated with
+    // `bun run db:generate && bun run db:migrate`. The framework's runtime migrator is
+    // not designed for serverless and adds ~2 min cold-start overhead per agent.
+    await runtime.initialize({ skipMigrations: true });
 
     // Wrap Babylon plugin BEFORE registering (so wrapped version is used)
     // This ensures all actions and provider accesses are logged when executed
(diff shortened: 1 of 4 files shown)


revlentless (approved):
This PR skips elizaOS's runtime migrations on all three `runtime.initialize()` calls and re-exports the plugin-sql schema tables so Drizzle Kit manages them instead. The approach is correct and well-documented. Confirmed that `skipMigrations` is a real option in `@elizaos/core`, all call sites are covered, and the `@elizaos/plugin-sql/node` subpath export is stable. Typecheck and biome pass clean.

Two minor suggestions (non-blocking):
- Consider pinning `@elizaos/plugin-sql` to an exact version rather than `^1.6.4` to avoid silent schema drift on install.
- A runtime guard in `eliza.ts` (`if (!plugin?.schema) throw ...`) would give a clearer error if the plugin restructures its exports in a future version.

No changes made during review.

LGTM

odilitime:
is elizaOS not ElizaOS!

odilitime (approved):

PR #1258, opened 2026-03-17
fix(db): remove elizaEmbeddingTable re-export that requires pgvector extension

revlentless:
## Summary

- `bun db:push` / `db:generate` fails with `type "vector" does not exist` because `eliza.ts` re-exports the ElizaOS `embeddingTable` which uses pgvector `vector` columns
- Removes the `elizaEmbeddingTable` re-export — no runtime or migration code references it, and ElizaOS creates the table at agent startup when needed

## Type

- [x] Bug fix

## Context / Links

- Regression introduced by PR #1248 (`fix: skip elizaOS runtime migrations, manage tables via Drizzle`)
- Related: PR #1254, PR #1255 (Vercel Lambda size fixes that touched the same area)

## Scope (keep it focused)

- [x] This PR is focused on a single change/theme (not a catch‑all)
- [x] Drive‑by refactors are excluded or split into a separate PR
- [x] Non‑goals / follow‑ups are listed below (with links)

**Areas touched**
- [x] DB (`packages/db`)

## Changes

- `packages/db/src/schema/eliza.ts` — removed `elizaEmbeddingTable` export (the only table using pgvector `vector` columns)

## Review guide

**Start here**
- `packages/db/src/schema/eliza.ts`

**Risk**
- [x] Low

**Notes for reviewers**
- Confirmed via grep: `elizaEmbeddingTable` is not referenced anywhere in the codebase (it was removed from the barrel export in PR #1255)
- The embeddings table is created by ElizaOS runtime when an agent needs it — no Drizzle migration needed

## Test plan

**Commands run**
- [x] `bun run typecheck`

**Manual verification**
1. Run `bun db:push` — should no longer error with "type vector does not exist"

- [x] No deploy impact

## Breaking changes

- [x] None

## Security / privacy

- [x] No security impact

- Reason: Backend-only DB schema config fix, no UI impact

## Checklist (author)
- [ ] Tests added/updated for behavior changes (or rationale provided)
- [ ] Dependency changes are intentional (`bun.lock` updated)

### DIFF
--- a/packages/db/src/schema/eliza.ts
+++ b/packages/db/src/schema/eliza.ts
@@ -17,7 +17,9 @@ export const elizaAgentTable = pluginSchema.agentTable;
 export const elizaRoomTable = pluginSchema.roomTable;
 export const elizaParticipantTable = pluginSchema.participantTable;
 export const elizaMemoryTable = pluginSchema.memoryTable;
-export const elizaEmbeddingTable = pluginSchema.embeddingTable;
+// NOTE: embeddingTable omitted — it uses pgvector `vector` columns which
+// require `CREATE EXTENSION vector` and break `db:push` / `db:generate` on
+// databases without the extension. ElizaOS runtime creates it when needed.
 export const elizaEntityTable = pluginSchema.entityTable;
 export const elizaRelationshipTable = pluginSchema.relationshipTable;
 export const elizaComponentTable = pluginSchema.componentTable;


SYMBaiEX (approved):
Approved. Clean, well-scoped fix — elizaEmbeddingTable should never have been in the Drizzle schema to begin with since ElizaOS owns that table's lifecycle (it creates it at runtime, conditional on the vector extension being available). Removing it unblocks db:push without any runtime regression; the symbol had no references outside eliza.ts and was already excluded from the barrel export since #1255.

One follow-up worth tracking separately: Neon production should have CREATE EXTENSION IF NOT EXISTS vector run once so ElizaOS can actually create the embeddings table at agent startup and agents get full semantic memory retrieval. Without it, searchMemoriesByEmbedding is silently broken in prod. Not a blocker for this PR.

PR #1265, opened 2026-03-18
fix(agent-chat): remove profanity filter from user input

SYMBaiEX:
## Summary

- Removes the `BLOCKED_PATTERNS` profanity check loop from `checkUserInput()` in `packages/shared/src/utils/content-safety.ts`
- Users were seeing a toast — *"Content contains inappropriate language"* — when sending messages containing common expletives to agents
- This is inconsistent with the platform's stated moderation policy (only bans scamming and CSAM)

## What stays

- Prompt injection detection (`INJECTION_PATTERNS`) — unchanged
- Spam / repetition detection — unchanged
- Message length limit (2000 chars) — unchanged
- `checkAgentOutput()` profanity guard — unchanged (agents generating expletives unprompted is a separate concern)

## Test plan

- [ ] Send a message containing an expletive in agent chat — should go through without a toast error
- [ ] Send a prompt injection attempt (`ignore previous instructions`) — should still be blocked
- [ ] Send a message > 2000 chars — should still be blocked

### DIFF
--- a/packages/shared/src/utils/content-safety.ts
+++ b/packages/shared/src/utils/content-safety.ts
@@ -16,11 +16,11 @@ const BLOCKED_PATTERNS = [
 
 // Prompt injection attempts
 const INJECTION_PATTERNS = [
-  /ignore (previous|all|above) instructions?/gi,
-  /system:?\s*you are now/gi,
-  /\[system\]/gi,
-  /forget (everything|all|previous)/gi,
-  /<\|.*?\|>/gi, // Special tokens
+  /ignore (previous|all|above) instructions?/i,
+  /system:?\s*you are now/i,
+  /\[system\]/i,
+  /forget (everything|all|previous)/i,
+  /<\|.*?\|>/i, // Special tokens
 ];
 
 export interface ContentCheckResult {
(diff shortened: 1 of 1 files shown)


slkzgm:
I reviewed the merged diff, the follow-up commits, and the existing review threads.

Main concern: this change is broader than the PR title/description suggests. `checkUserInput()` no longer removes only profanity. It also stops blocking the non-profanity entries currently bundled into `BLOCKED_PATTERNS`, including `kill yourself` / `suicide` and `hack` / `exploit` / `ddos` in `packages/shared/src/utils/content-safety.ts`.

If the intended policy change is specifically "allow expletives in user chat", I do not think removing the entire loop is the right implementation. The safer and clearer version would be to split the patterns by concern and keep only the profanity subset out of `checkUserInput()`, or explicitly document that user input is now allowed to contain self-harm / attack-related phrases as well.

Secondary issue: `BLOCKED_PATTERNS` still uses `/gi` and is evaluated with `.test()` in `checkAgentOutput()`. That makes the filter stateful across calls because of `lastIndex`, so repeated checks can alternate between block / allow for the same content. The PR already fixed this exact class of bug for `INJECTION_PATTERNS`, but the same problem still exists for agent output filtering.

I would treat this as a follow-up before calling the area stable:
1. Split `BLOCKED_PATTERNS` by concern and make the user-input policy explicit.
2. Remove the `g` flag from the remaining `.test()`-based blocklists (or reset `lastIndex`).
3. Add regression tests covering the new allowed-user-input behavior and repeated agent-output checks.

PR #1264, opened 2026-03-18
feat: BAB-286 notification system UI — win/loss pop-ups, feed cards, settings

jasperdisney:
## Summary

Full frontend implementation of the V1 notification system (BAB-286), covering all three tiers defined in BAB-277.

### Tier 1 — Win/Loss Outcome Pop-ups
- Full-screen overlay with canvas-based confetti burst animation (32 waves, 400 particles/wave) for wins
- Animated points counter with scale-up effect and digit-stable width
- Agent attribution by name ("via your agent **Ares**") — never generic
- Loss pop-up uses intentionally softer treatment: no confetti, no glow, drops from above (loss aversion design principle)
- Auto-dismisses after 5 seconds with visible progress bar, or on tap
- Card tap navigates to market deep link, backdrop tap dismisses without navigation
- Sequential queue for individual notifications
- Stacked summary popup ("X markets resolved while you were away") for batch notifications with scrollable results list, win/loss counts, net points, and confetti on net-positive

### Tier 3 — Feed Signal Cards (visual design only)
- `MarketClosingSoonCard` — position side badge, PnL%, time remaining
- `TopGainerCard` — points gained, percentage, agent attribution
- `TopLoserCard` — points lost, percentage, agent attribution
- Cards are full-width tappable links with accent-colored hover states
- Consistent with existing feed card visual language (not system alerts)

### Tier 2 — Notifications Settings Tab
- Performance digest toggle (on/off) with frequency selector (hourly/daily/weekly, default: daily)
- Delivery channel selector (in-app/email/both, default: both)
- Tier 1 and Tier 3 intentionally not shown in settings per spec
- Settings gear icon added to `/notifications` page header → links to `/settings?tab=notifications`
- State is local (`useState`) — backend API wiring is a TODO

### Other UI Tweaks
- Removed icon from "Open Positions" sidebar title
- Matched "View Positions" button style to "View Full Portfolio"
- Cleaned up Settings tabs (Security, Privacy, API Keys): removed redundant headers, block icons, Active Session block, Account Privacy & Deletion block

## Important — Do Not Merge Directly

This PR contains **frontend only**. Before merging to production, the following must be completed:

1. **Backend** (Patrick / Luca): market resolution event hooks, notification queue storage, digest scheduler, notification settings API
2. **Trigger wiring** (Brandon / BAB-285): wire SSE/websocket events to `showOutcome()` / `showBatchOutcomes()`, inject feed signal cards into the feed, implement capping/dedup logic
[...]

PR #1267, opened 2026-03-18
fix: viewport-aware Tooltip component, migrate OnChainBadge

jasperdisney:
## Summary

- Add a shared `Tooltip` component (`components/ui/tooltip.tsx`) built on `@radix-ui/react-tooltip` (already a dependency, previously unused)
- Migrate `OnChainBadge` from a hand-rolled absolute-positioned tooltip to the shared component

## Problem

Tooltips positioned with `absolute + left-1/2 + -translate-x-1/2` get clipped when near viewport edges. For example, the "Verified On-Chain" badge tooltip on profile pages clips behind the left sidebar when the user's display name is short, pushing the badge close to the sidebar edge. The same issue affects the leaderboard page.

## Solution

The shared `Tooltip` component solves this by:
- **Portal rendering** — tooltip appends to `document.body`, escaping any `overflow-hidden` ancestor (like the sidebar)
- **Collision detection** — `collisionPadding={8}` shifts the tooltip to stay within viewport bounds
- **Auto-flipping** — preferred `side` is a hint; Radix flips to the opposite side when space is insufficient

## Usage

```tsx
import { Tooltip } from '@/components/ui/tooltip';

<Tooltip content="Simple text tooltip">
  <button>Hover me</button>
</Tooltip>

<Tooltip
  content={<div><p>Rich content</p></div>}
  side="right"
  align="start"
>
  <span>Hover me</span>
</Tooltip>
```

## Scope

This PR migrates `OnChainBadge` only (highest risk — appears on profile pages and leaderboard near sidebar edge). Other components with the same clipping issue (`GameOnboardingTooltip`, `LikeButton` reaction picker, `Dropdown`) are candidates for follow-up PRs — they require different approaches since they're interactive popups, not simple hover tooltips.

## Files Changed
- `apps/web/src/components/ui/tooltip.tsx` — new shared component
- `apps/web/src/components/profile/OnChainBadge.tsx` — migrated from hand-rolled tooltip to shared component, removed manual `useState` hover state

### DIFF
--- a/apps/web/src/components/profile/OnChainBadge.tsx
+++ b/apps/web/src/components/profile/OnChainBadge.tsx
@@ -2,35 +2,8 @@
 
 import { cn } from '@babylon/shared';
 import { Shield, ShieldCheck } from 'lucide-react';
-import { useState } from 'react';
+import { Tooltip } from '@/components/ui/tooltip';
 
-/**
- * On-chain badge component for displaying blockchain verification status.
- *
- * Displays a badge indicating whether a user's identity is verified on-chain
- * via NFT registration. Shows different icons and tooltips for verified vs
- * unverified states. Includes hover tooltip with NFT token ID for verified users.
- *
- * Features:
- * - Verified/unverified state display
- * - NFT token ID display in tooltip
- * - Size variants (sm, md, lg)
- * - Optional label text
- * - Hover tooltips
- *
- * @param props - OnChainBadge component props
- * @returns On-chain badge element
- *
- * @example
- * ```tsx
- * <OnChainBadge
- *   isRegistered={true}
- *   nftTokenId={123}
- *   size="md"
- *   showLabel={true}
- * />
- * ```
- */
 interface OnChainBadgeProps {
   isRegistered: boolean;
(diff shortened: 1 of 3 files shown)


SYMBaiEX (approved):
**Approved** ✅

Clean, focused implementation — uses the already-installed `@radix-ui/react-tooltip` to solve a real viewport clipping problem with portal rendering + collision detection. `OnChainBadge` migration removes ~90 lines of manual hover state management and the tooltip now auto-repositions near sidebar/viewport edges.

One improvement applied before merge: lifted `TooltipProvider` from per-instance to the app root (`Providers.tsx`, both branches) so all tooltips share the same hover-intent delay state — the standard Radix pattern for multi-tooltip apps. Per-instance Providers create independent delay timers, causing the delay to re-trigger when quickly moving between adjacent tooltips.

Also rebased onto current `production` to pick up the `OutcomeNotificationProvider` from #1264 — no conflicts in final state.

✅ TypeScript clean | ✅ Biome clean | ✅ Lint zero warnings | ✅ Rebased on production

PR #1268, opened 2026-03-18
feat(feed): replace Hot tab with Stories tab (BAB-134, BAB-199)

SYMBaiEX:
## Summary

- Replaces the **Hot** feed tab with a **Stories** tab that surfaces today's daily topic prominently
- New `/api/feed/stories` API: narrative-scored stories with 2× score boost for stories matching today's `topicKey` from `DailyTopicService`
- New `DailyTopicBanner` component shown at the top of the Stories tab when a daily topic exists
- Reuses existing `ForYouFeedList` renderer and `NarrativeStory` type — no new rendering logic
- `FeedSurface` analytics union updated with `'stories'` (keeps `'hot'` for backward-compat event data)

## Commits

1. `feat(shared)`: add `'stories'` to `FeedSurface` union
2. `feat(feed)`: add `/api/feed/stories` route with daily-topic boost scoring
3. `feat(feed)`: add `useStoriesFeed` paginated SSE-aware hook
4. `feat(feed)`: add `DailyTopicBanner` component and stories `EmptyFeed` variant
5. `feat(feed)`: replace Hot tab with Stories tab wiring in `FeedClient` and `FeedToggle`

## Test plan

- [ ] Stories tab appears, Hot tab is gone
- [ ] `DailyTopicBanner` shows today's topic label and summary at the top of the Stories tab
- [ ] Stories are ranked with topic-matching stories first, then others by narrative score
- [ ] Empty state ("No Stories Yet") renders when no stories exist
- [ ] Pull-to-refresh works on Stories tab
- [ ] SSE `feed` events trigger Stories refresh (confirm via network tab — 2s debounce)
- [ ] For You, Latest, Following, Trades tabs completely unaffected
- [ ] `bun run typecheck` passes (zero errors in changed files)
- [ ] `bun run lint` passes (zero warnings)


slkzgm (changes requested):
I'm requesting changes before merge. The Stories tab direction makes sense, but there are a few blocking issues in the current implementation:

- Stories analytics are broken in two places. The backend validator still rejects `surface: "stories"` in `/api/feed/events`, and the shared renderer still hardcodes `surface: "for_you"` when emitting feed events. In practice that means Stories events are either rejected or misattributed.
- `useStoriesFeed` is implemented as a paginated hook (`offset`, `limit`, `hasMore`, `loadMore`), but `/api/feed/stories` ignores those params and always returns the full array. The hook and route do not share the same contract right now.
- `DailyTopicBanner` is not actually shown whenever a daily topic exists. `FeedClient` returns the Stories empty state before rendering the banner, so an empty Stories tab hides the daily topic even though the PR description says the banner should sit at the top of the tab when a daily topic exists.
- The new Stories route duplicates a large amount of the narrative pipeline inline inside an app route. Given the repo rule that apps should stay wiring-only, plus the explicit DRY requirement for this review, I don't think we should merge another 600+ lines of near-copy without extracting the shared story-building logic and adding focused tests around the new topic-boost behaviour.

Once those are addressed, I'm happy to take another pass.

PR #1282, opened 2026-03-19
fix(feed): deep-clone cached stories before per-user enrichment

SYMBaiEX:
## Summary

Fixes the critical cache mutation race condition identified in the PR #1268 review (see [comment](https://github.com/BabylonSocial/babylon/pull/1268#issuecomment-4088787516)).

- `getCacheOrFetch` returns the **same in-memory object** for all requests within the 60s TTL window
- `enrichStoriesForUser` was mutating `story.posts[].isLiked` and `isShared` directly on that shared reference
- This caused user A's like/share state to bleed into user B's response, and created race conditions under concurrent load
- Fix: `structuredClone(fullResult.stories)` before enrichment so each request operates on its own copy

## Test plan

- [ ] Two different authenticated users see their own like/share state (not each other's)
- [ ] Unauthenticated requests are unaffected (no enrichment path)
- [ ] Pagination still works correctly on the cloned array


SYMBaiEX:
## Review feedback addressed

The core fix is correct and well-scoped. `structuredClone` before `enrichStoriesForUser` prevents the shared cached object from leaking one user's `isLiked`/`isShared` state into another's response.

The latest commit adds:
- **Regression test** (`route.test.ts`) — 4 tests proving the cache object is never mutated, two users see independent enrichment state, anonymous users skip enrichment, and pagination works on the cloned array.
- **Performance optimization** — anonymous requests now skip the `structuredClone` since they never hit the enrichment path.

Note: the Copilot review comments about system-status-service, solana-send-transaction, content-safety, digest-settings, messaging schema, and notification-service are all on files **not changed by this PR** (the reviewer compared against the wrong base, seeing 201 files instead of 3). Those are pre-existing issues worth addressing in separate PRs.

PR #1286, opened 2026-03-19
fix(db): make achievement migration idempotent for prod

revlentless:
## Summary

- The `0045_spicy_blindfold` migration was a squashed bundle from the original achievements branch that included DDL for tables, enums, and columns already created by prior migrations on production (`SentryWebhookInboxStatus`, `MessageReaction`, `SentryWebhookInbox`, `WalletTransfer*`, `User` email columns)
- This caused `db:migrate` to fail with `type "SentryWebhookInboxStatus" already exists`
- The journal `when` timestamp was also older than already-applied migrations, so Drizzle skipped it

## Fix

- Stripped the migration down to only the 4 achievement/challenge tables with `IF NOT EXISTS` guards
- Updated journal timestamp to `1773940000000` (newer than all existing migrations)

## Verification

- `bun run db:migrate` ran successfully against prod
- `bun run scripts/seed-achievements.ts` seeded 15 achievements + 40 challenges
- Tables confirmed present: `AchievementDefinition`, `ChallengeDefinition`, `UserAchievement`, `UserChallengeProgress`


SYMBaiEX (approved):
Approved — verified merge readiness:

- **Migration correctness**: All 18 removed DDL statements confirmed present in dedicated migrations (0040–0045). Only the 4 achievement/challenge tables + 3 indexes remain, which are not defined elsewhere.
- **Idempotency**: `IF NOT EXISTS` guards on all `CREATE TABLE` and `CREATE INDEX` statements.
- **Journal timestamp**: `1773940000000` correctly orders after idx:52 (`1773853200000`).
- **CI**: All checks passing.
- **No unresolved review comments**.

PR #1297, opened 2026-03-20
fix: exclude zero-share positions from market resolution notifications

revlentless:
## Summary

- **Fix**: Market resolution notifications included fully-sold positions (shares=0) with pnl=0, causing won markets to show "+0 pts" and net totals to be skewed negative
- **Root cause**: `notifyResolvedMarketOwners` queried all resolved positions without filtering out zero-share ones. Positions sold before resolution had `pnl = payout - costBasis = 0 - 0 = 0`.
- **Fix**: Add `shares > 0` filter to the notification query WHERE clause

## Type

- [x] Bug fix

## Context / Links

- [BAB-323](https://linear.app/eliza-labs/issue/BAB-323/bug-market-resolution-notification-shows-wrong-pandl-won-markets)
- Related: BF-94, BF-50 (prior P&L calculation bugs)

## Scope (keep it focused)

**Areas touched**

**Non-goals / follow-ups:**
- BAB-324: Wallet data inconsistencies (sidebar inflation, P&L ±100% formula, ghost positions) — separate PR

## Changes

- `apps/web/src/lib/services/market-resolution-notifications.ts`: Added `gt(positions.shares, '0')` to the notification query WHERE clause + imported `gt` from `@babylon/db`
- `packages/testing/unit/market-resolution-notifications.test.ts`: Added 3 test cases for edge cases (zero-point, negative-point, mixed YES/NO aggregation)

## Review guide

**Start here**
- `apps/web/src/lib/services/market-resolution-notifications.ts` — the one-line fix in the WHERE clause (line ~110)

**Risk**
- [x] Low

**Notes for reviewers**
- The resolution code (`PredictionMarketService.resolve()`) intentionally continues processing all positions including closed ones for accounting completeness — this fix only filters at the notification layer
- Zero-share positions are those fully sold before resolution; their pnl is always 0 at resolution time since `payout = shares = 0` and `costBasis = avgPrice * shares = 0`
- The `outcome` derivation (`points >= 0 ? 'win' : 'loss'`) remains correct once zero-share positions are excluded

## Test plan

**Commands run**
- [x] `bun run typecheck`
- [x] `bun run lint`

**Manual verification**
1. After deploying, trigger market resolution on a market where an agent sold positions before resolution — the "while you were away" notification should only include positions with shares > 0
2. Won markets should show positive point values; lost markets should show negative
3. Net total should equal sum of individual P&Ls

- [x] No deploy impact

## Breaking changes

- [x] None

## Security / privacy

- [x] No security impact

- Reason: Backend-only query filter change. The notification UI components are unchanged — they will now receive correct data. No visual verification possible without live market resolution events.

## Checklist (author)

### DIFF
--- a/apps/web/src/lib/services/market-resolution-notifications.ts
+++ b/apps/web/src/lib/services/market-resolution-notifications.ts
@@ -3,6 +3,7 @@ import {
   and,
   db,
   eq,
+  gt,
   isNotNull,
   type JsonValue,
   markets,
@@ -106,7 +107,8 @@ export async function notifyResolvedMarketOwners(
         eq(positions.status, 'resolved'),
         isNotNull(positions.outcome),
         isNotNull(positions.pnl),
-        isNotNull(positions.resolvedAt)
+        isNotNull(positions.resolvedAt),
+        gt(positions.shares, '0')
       )
     );
 
--- a/apps/web/src/lib/services/market-resolution-notifications.ts
+++ b/apps/web/src/lib/services/market-resolution-notifications.ts
@@ -19,6 +19,8 @@ interface ResolvedOutcomeRow {
   marketName: string;
   points: number;
   agentName: string | null;
+  /** Position outcome from DB (true = winning side). */
+  positionWon: boolean;
 }
 
 export interface GroupedResolvedOutcome {
@@ -42,15 +44,20 @@ function formatPoints(points: number): string {
 export function groupResolvedMarketOutcomes(
   rows: ResolvedOutcomeRow[]
 ): GroupedResolvedOutcome[] {
-  // outcome is intentionally omitted during accumulation — points change as
-  // rows are merged, so outcome is derived once at the end from final points.
-  const grouped = new Map<string, Omit<GroupedResolvedOutcome, 'outcome'>>();
+  // Track both accumulated points and whether any row was on the winning side.
+  // A holder may have YES and NO positions on the same market — the final
+  // outcome is 'win' when at least one position resolved on the winning side.
+  const grouped = new Map<
+    string,
+    Omit<GroupedResolvedOutcome, 'outcome'> & { hasWinningPosition: boolean }
+  >();
 
   for (const row of rows) {
     const key = `${row.ownerUserId}:${row.holderId}:${row.marketId}`;
     const existing = grouped.get(key);
     if (existing) {
       existing.points = Number((existing.points + row.points).toFixed(2));
+      if (row.positionWon) existing.hasWinningPosition = true;
       continue;
     }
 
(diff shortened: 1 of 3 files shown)


slkzgm (changes requested):
I pushed a low-risk follow-up commit (`f3954bdbc`) adding the missing `gt(positions.shares, "0")` filter to `notification-digest-service.ts`, so zero-share positions are now excluded from both immediate notifications and digests.

I am still requesting changes for one remaining correctness issue:

- `apps/web/src/lib/services/market-resolution-notifications.ts:47-80` now classifies the grouped result as `win` whenever any row was on the winning side. That fixes the single-position `avgPrice > 1` case, but it regresses holders who have both YES and NO exposure on the same market.
- Mixed-side holdings are explicitly supported today: `PredictionMarketService.sell()` loads both the YES and NO positions and throws `"Multiple positions exist on this market. Specify positionId."` when both are open (`packages/core/markets/prediction/PredictionMarketService.ts:232-250`). `PredictionDbAdapter.getPosition()` is also side-specific (`packages/core/markets/prediction/adapters/drizzle/PredictionDbAdapter.ts:245-269`).
- With the new grouping rule, a holder can end up with negative net points but still be labeled `outcome: "win"`. `buildMessage()` then renders `"won X points"` because it combines the boolean outcome with `Math.abs(points)`. The new test `packages/testing/unit/market-resolution-notifications.test.ts` codifies that behavior for a `-800` points example.

I do not think a grouped market-level notification can safely collapse mixed YES/NO exposure into a single boolean outcome. This needs either split notifications per side/position, or grouped messaging that stays tied to net PnL while exposing winner-side information separately.

slkzgm (changes requested):
The zero-share filter looks good, but the new grouped outcome rule still regresses mixed YES/NO holders.

`groupResolvedMarketOutcomes()` now marks a grouped market as `win` whenever any underlying row resolved on the winning side. Mixed-side exposure is a real supported case in the current domain model: `PredictionMarketService.sell()` explicitly handles both YES and NO positions on the same market and asks for `positionId` when both are open, and `PredictionDbAdapter.getPosition()` is side-specific.

With the current implementation, a holder can end up with negative net points but still receive `outcome: "win"`. That flows directly into the popup and digest UX:
- popup styling/confetti is keyed from `outcome === "win"`
- the stacked summary colors/counts are keyed from `outcome`
- the digest now reports `marketsWon/marketsLost` from `outcome`
- the message body still combines that boolean with `Math.abs(points)`

So a user can see a green "win" treatment for a `-800` result, and the digest can say `-800 points net | 1 won | 0 lost`.

I do not think a market-level notification can safely collapse mixed YES/NO exposure into a single win/loss boolean by OR-ing position outcomes. Either keep the grouped outcome tied to the final net points, or stop collapsing mixed-side exposure into a single boolean notification.

slkzgm:
Closing this PR because the remaining BAB-323 work needs to be reworked as a separate implementation pass.

The zero-share filter change is valid, but the broader notification semantics for mixed YES/NO exposure still need a cleaner end-to-end solution. Rather than keep iterating in this PR, we will handle the related BAB-323 issue separately in a fresh follow-up.

Thanks for the work here.

PR #1298, opened 2026-03-20
fix(web): distribute markets evenly, add content backfill, show composer on all feed tabs

SYMBaiEX:
## Summary

- **Markets distribution**: New `distributeMarkets()` replaces `spreadNewMarkets()` for true even-interval market card placement throughout For You and Stories feeds (was clustering at bottom)
- **Stories market cards**: Ported the new-market query block into the Stories pipeline (was entirely missing — Stories had zero market cards)
- **Stories backfill (12h→7d)**: When primary 12h content is sparse, fills remaining capacity with older posts relevant to today's daily topic via `questions.topicKey` join + `isTextOnTopic()` keyword filter
- **For You backfill (24h→7d)**: When primary 24h content is sparse, fills remaining capacity with high-engagement posts from the past week ordered by hot score (likes + comments×2 + shares×3)
- **InlineComposer**: Now shows on For You and Stories tabs for authenticated users (was latest-only)

## Production safety

| Concern | Impact |
|---------|--------|
| DB query cost | Backfill queries only fire when primary < 500 (low-activity periods). Same LIMIT cap. |
| Cache keys | Identical keys and TTLs — no multiplication |
| Serverless cold starts | No new API routes or params — same function signatures |
| Memory per request | Same 500-item pipeline cap — backfill fills the gap, doesn't add on top |
| Content quality | Backfill scored through same ranking pipeline — not random, not slop |

## Files changed

| File | Change |
|------|--------|
| `apps/web/src/app/api/feed/for-you/scoring.ts` | Add `distributeMarkets()` |
| `apps/web/src/app/api/feed/for-you/pipeline.ts` | Switch to `distributeMarkets()` + hot-post backfill |
| `apps/web/src/app/api/feed/stories/pipeline.ts` | Market cards + `distributeMarkets()` + topic-relevant backfill |
| `apps/web/src/app/feed/FeedClient.tsx` | Extend InlineComposer tab condition |

## Test plan

- [ ] For You feed first page (20 items) contains 2-4 distributed market cards (not clustered at bottom)
- [ ] Stories feed includes market cards interspersed with posts
- [ ] Stories feed shows topic-relevant older content when <12h content is sparse
- [ ] For You feed shows engagement-ranked older content when <24h content is sparse
- [ ] "You're all caught up" only appears when 7-day backfill is also exhausted
- [ ] InlineComposer appears on For You and Stories tabs for authenticated users
- [ ] Posts created from For You/Stories tabs appear in Latest feed


slkzgm (changes requested):
Thanks for the detailed PR description. I think the overall direction is right, but I'm requesting changes for two correctness issues and one performance / reuse concern.

1. Stories new-market cards currently build `anchorPostId`, but the card is returned with `posts: []` and `postIds` is built only from `story.posts`. In practice that means the `NewMarketCard` interaction bar falls back to zero counts and never gets per-user like/share state in Stories, because the route only enriches `fullResult.postIds`. This is a real regression compared with the existing For You path, where the anchor post is hydrated into the story payload.

2. The Stories market query uses the 12h feed cutoff instead of the 24h new-market window used elsewhere. As written, Stories will miss markets opened 12-24 hours ago, so it does not actually match the PR description of porting the existing new-market behavior into Stories.

3. The For You backfill ordering recomputes engagement with correlated `COUNT(*)` subqueries per row even though the repo already has `mv_post_interaction_counts` / `engagement_score`. That duplicates existing logic and adds a potentially expensive sort path exactly in the sparse-feed case where this code is supposed to help. I'd strongly prefer reusing the existing materialized-view path or at least a single aggregate join/CTE.

I didn't see test coverage added for the new `distributeMarkets()` behavior or for Stories new-market hydration, and the current targeted tests still all pass without exercising these paths. Once the correctness issues are addressed, I'm happy to take another pass.

PR #1301, opened 2026-03-20
fix(engine): de-emphasize crypto content from stories feed

SYMBaiEX:
## Summary
- Adds a `TOPIC_BLOCKLIST` to `DailyTopicService` that filters out 28 crypto-related keywords (bitcoin, ethereum, defi, nft, etc.) from daily topic candidate selection — prevents crypto from appearing as "Today's Story" in the banner
- Removes crypto story templates (9 headlines) and 1 breaking news template from `StorySeedService`
- Removes 2 crypto story seeds from `TopicDiversityService`

The `crypto` editorial beat is kept intact so the system remains structurally sound — RSS feeds still ingest crypto headlines for parody generation and other systems, but crypto content no longer surfaces in the user-facing Stories tab.

## Test plan
- [ ] Verify `bun run typecheck` passes for `@babylon/engine`
- [ ] Verify `bun run build` succeeds
- [ ] Confirm "Today's Story" banner no longer shows crypto topics (Bitcoin, Ethereum, etc.)
- [ ] Confirm non-crypto topics (tech, politics, finance, AI) still appear normally

### DIFF
--- a/packages/engine/src/services/daily-topic-service.ts
+++ b/packages/engine/src/services/daily-topic-service.ts
@@ -12,6 +12,37 @@ import {
 } from '@babylon/db';
 import { logger } from '@babylon/shared';
 
+/**
+ * Topics that should never be selected as the daily topic.
+ * Prevents crypto-native content from dominating the "Today's Story" banner,
+ * keeping the simulation feeling like a broad real-world news experience.
+ */
+const TOPIC_BLOCKLIST = new Set([
+  'bitcoin',
+  'btc',
+  'ethereum',
+  'eth',
+  'crypto',
+  'cryptocurrency',
+  'blockchain',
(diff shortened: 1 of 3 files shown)


revlentless:
Looks good - clean and well-scoped. 
  
  One thing: extractTopicTokens filters to token.length >= 4, so btc, eth, and nft in the blocklist are dead code - they'll
  never appear as candidate tokens. Not a blocker, but either remove them (with a comment explaining the min-length) or keep them with a // these are <4 chars  
  and currently filtered by extractTopicTokens, included for documentation note so nobody assumes they're actually working. Also token/tokens/mining are a bit
  generic - fine for now but could false-positive on non-crypto headlines down the road.

PR #1305, opened 2026-03-20
feat(web): redesign wallet page — balance, P&L, and positions tabs

jasperdisney:
## Summary

- Redesign all three wallet tabs (Balance, P&L, Positions) with card-based layouts, responsive mobile padding, dark-theme-friendly badges, and improved visual hierarchy
- Balance tab: hero total card, Cash/Open Positions breakdown, Agents Total row, member cards with agent badges
- P&L tab: summary cards with percentage change, bordered chart, clickable entity selector rows, click-outside dropdown, dynamic Y-axis width
- Positions tab: 2-row card layout (ticker+PnL / labeled data columns+action), leverage in side badge, CSS line-clamp for prediction questions, section counts
- Hide NFT promo banner on /wallet, fix mobile header spacing, tab bar height consistency
- Add dev mock data system (disabled via flag, zero bundle cost in production — all mock data behind dynamic imports)


slkzgm (changes requested):
I’m requesting changes for three concrete issues.

1. The PR currently commits a temporary dev-only mock/auth-bypass path into a PR that targets `production` (`apps/web/src/app/wallet/page.tsx`, `apps/web/src/components/wallet/__dev__/*`, `WALLET-DEV-REMINDER.md`). That adds a manual “remove before production” workflow, overrides Zustand fetchers with no-ops, and introduces an auth-bypass code path in production-bound files. We should not merge temporary harnesses/fallbacks like this into the main codepath.

2. The new P&L summary percentages are not mathematically meaningful right now. In `apps/web/src/components/wallet/v2/pnl-tab.tsx`, the Lifetime card divides `lifetimePnl` by itself, so it always renders `+100.0%` / `-100.0%`, and the `|| 1` fallback on the other cards produces fake percentages when lifetime P&L is zero. If we don’t have a real denominator for these values, we should omit the percentage rather than show misleading numbers.

3. The clickable member rows in `apps/web/src/components/wallet/v2/pnl-tab.tsx` are implemented as `div onClick`, so they are not keyboard accessible. Since these rows are now part of the entity-selection UX, they should be actual buttons (or equivalent accessible controls), not mouse-only containers.

slkzgm (approved):
Follow-up review after the latest fixes.

The remaining blockers I raised are now addressed:
- the temporary wallet mock/auth-bypass path is gone from the PR branch
- the misleading P&L percentages were removed from the summary cards
- the clickable P&L member rows are now keyboard-accessible buttons
- the dropdown outside-click behavior now reuses the existing shared hook

I only ran targeted file-level checks in this environment (Biome on the touched files). I could not run a full local typecheck here because the worktree does not have the repo's full type dependencies installed.

slkzgm (changes requested):
Requesting changes for two issues that still look material in the current diff.

1. The new Balance / P&L member cards are still built from partial local aggregates instead of the existing team summary source of truth. `apps/web/src/components/wallet/v2/balance-tab.tsx` hardcodes every agent's `cash` to `0` and derives totals from open positions only, while `apps/web/src/components/wallet/v2/pnl-tab.tsx` sets each agent row to `currentPnl = lifetimePnl = unrealized`. That means the redesigned UI can show materially wrong numbers for agent members and team totals whenever agents have idle balance or realized P&L. We already have `useTeamTradingSummary` (`apps/web/src/hooks/useTeamTradingSummary.ts`) that fetches owner + agent balances and lifetime P&L and computes `currentPnL` from those stable values. This PR should reuse that instead of re-deriving a second, less accurate model in the UI layer.

2. `apps/web/src/components/wallet/v2/pnl-chart.tsx` removes focus outlines from the entire chart subtree with `[&_*:focus]:outline-none [&_*]:outline-none`. That regresses keyboard accessibility without adding a replacement focus-visible treatment. If the chart or its descendants can receive focus, we should preserve a visible focus state rather than globally suppressing it.

Once the wallet tabs are backed by the real team summary data and the chart focus regression is fixed, I'm good to re-review.

slkzgm (approved):
Latest follow-up looks good to me.

The wallet tabs now consume the existing team summary source of truth for member cash / lifetime / current P&L, and the chart no longer strips focus outlines from the subtree. That addresses the two material issues from my last review.

slkzgm:
Acknowledged the latest review feedback and addressed it in `d751742d1`.

Changes in this follow-up:
- wired the wallet page to `useTeamTradingSummary` once at the page level and passed that source of truth into the Balance and P&L tabs
- stopped re-deriving agent cash / lifetime / current P&L inside the UI tabs from partial local aggregates
- removed the chart-wide focus outline suppression so keyboard focus is no longer globally hidden

Targeted Biome checks on the touched files passed. I still could not run a full local typecheck in this worktree because the repo dependencies/types are not fully installed here.

slkzgm (changes requested):
Requesting changes for one material regression in the latest follow-up.

1. The wallet page now renders the Balance and P&L tabs from `useTeamTradingSummary()` at the page level (`apps/web/src/app/wallet/page.tsx`), but that hook only fetches on mount / identity changes (`apps/web/src/hooks/useTeamTradingSummary.ts`). It is not wired into the existing wallet polling or invalidation paths. Meanwhile the page still polls `useWalletBalancePolling()` and `useUserPositionsPolling()`, and `PositionsTab` invalidates those stores after close/sell actions. The result is that Positions can update immediately after a trade or poll tick while the Balance / P&L member cards keep showing stale teamSummary data until a full page reload.

This needs to be tied back into the same refresh lifecycle before merge, either by deriving the displayed data from the already-polled stores or by explicitly refreshing `useTeamTradingSummary` on the same cadence / invalidation events. As a secondary perf issue, the new summary fetch also runs eagerly even though the default tab is `positions`, so we are paying for the extra balance/positions/agents requests before the user ever opens Balance or P&L.

slkzgm (approved):
Latest follow-up addresses the regression I called out.

`useTeamTradingSummary()` is no longer fetched eagerly on the default `positions` tab, and the Balance / P&L summary data is now refreshed off the same wallet balance / positions lifecycle instead of staying stale after polling or trade-driven invalidation.

slkzgm:
Acknowledged the latest review feedback and addressed it in `bc9f89254`.

What changed in this follow-up:
- `useTeamTradingSummary()` is now enabled only when the active tab is `balance` or `pnl`, so we no longer pay the extra summary requests while the default `positions` tab is shown
- the wallet page now watches the existing wallet balance / positions store refresh timestamps and calls `teamSummary.refresh()` when those already-polled stores update while Balance or P&L is active

That ties the team summary back into the same refresh lifecycle as the rest of the wallet page without expanding the tab logic again.

I re-ran a targeted Biome check on the touched file successfully.

slkzgm (dismissed):
Latest follow-up addresses the remaining concern for me.

`useTeamTradingSummary()` is now only enabled on the Balance / P&L tabs, so we no longer pay the extra summary fetch cost on the default Positions view, and the page refreshes the summary off the same wallet balance / positions store lifecycle instead of leaving Balance / P&L stale after polling or trade-driven invalidation.

I re-ran targeted checks on the touched file in this environment and did not find any new blockers in the current diff.

PR #1334, opened 2026-03-22
fix: add privyId→PK fallback for did:privy users stored as id

SYMBaiEX:
## Summary

- Fixes production `duplicate key violates unique constraint "User_pkey"` errors on `GET /api/users/[userId]/portfolio-breakdown` and `balance` routes
- After PR #1327's classification-based routing, `did:privy:…` identifiers are classified as `privyId` and only query `users.privyId` — missing users where that value is stored as `users.id` (primary key)
- Adds a PK fallback for `privyId`-classified lookups that miss, keeping single-column indexed queries (no OR reversion)
- Applied in both `packages/api` lookup layer and `packages/engine` portfolio breakdown

## Root cause

PR #1327 optimized `findUserByIdentifier` from `WHERE id = $1 OR privyId = $1 OR lower(username) = lower($1)` to classification-based single-column routing. Some users have their `did:privy:…` value stored as `users.id` (not `users.privyId`). The classifier routes these to `WHERE privyId = $1` which misses → null → inline INSERT → PK violation.

## Fix

When a `privyId`-classified lookup returns no rows, fall back to `WHERE id = $1`. Both queries hit single-column indexes (PK is O(1)), preserving the performance optimization.

## Test plan

- [ ] Verify `did:privy:…` users stored as `users.id` are found by `findUserByIdentifier`
- [ ] Verify `GET /api/users/{did:privy:...}/portfolio-breakdown` returns data (not 500)
- [ ] Verify `GET /api/users/{did:privy:...}/balance` returns data (not 500)
- [ ] Verify normal UUID/username lookups are unaffected
- [ ] Verify no duplicate key errors in production logs

### DIFF
--- a/packages/api/src/cache/cached-database-service.ts
+++ b/packages/api/src/cache/cached-database-service.ts
@@ -38,7 +38,7 @@ import {
   users,
 } from '@babylon/db';
 import { StaticDataRegistry } from '@babylon/engine';
-import { logger } from '@babylon/shared';
+import { logger, resolveUserIdentifierKind } from '@babylon/shared';
 import {
   CACHE_KEYS,
   DEFAULT_TTLS,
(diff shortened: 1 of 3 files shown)


SYMBaiEX:
**Re: Copilot review on cache invalidation**

Good catch — fixed in 2e4aa62. `invalidateUserIdentifierCaches` now uses `resolveUserIdentifierKind(user.id)` to detect when `user.id` is a `did:privy:…` value and invalidates `privy:${user.id}` alongside the standard keys. This covers the case where `user.privyId` is null but lookups cached under the `privy:` prefix via the fallback path.

PR #1336, opened 2026-03-23
prediction trading errors return 500 instead of proper error message

millw14:
PredictionMarketService threw plain Error objects for all business logic errors (market resolved, expired, insufficient shares, etc.). The error handler only recognized BabylonError subclasses, so in production all prediction errors returned HTTP 500 with a generic message instead of the actual error.

- Replace plain Error throws with BadRequestError/NotFoundError from @babylon/shared

- Update error handler to recognize @babylon/shared BabylonError instances


revlentless:
Verified the fix is correct and all existing test assertions still match.

LGTM!

PR #1358, opened 2026-03-26
chore: request ElizaReview in CODEOWNERS

slkzgm:
## Summary
- add `@ElizaReview` to every effective CODEOWNERS rule
- ensure the review bot is still requested when more specific path rules match

## Test plan
- not run (repository configuration change only)

## Screenshots / recordings
- Reason: repository configuration change, no UI impact

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,53 +7,53 @@
 # - Keep this file coarse-grained. It routes reviews; it does not define who may edit code.
 
 # Fallback
-* @slkzgm @SYMBaiEX
+* @slkzgm @SYMBaiEX @ElizaReview
 
 # Apps
-/apps/cli/ @slkzgm @odilitime
-/apps/docs/ @slkzgm @odilitime
-/apps/web/ @slkzgm @SYMBaiEX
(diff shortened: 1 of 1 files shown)


slkzgm (commented):
Reviewed the CODEOWNERS change end-to-end. No implementation findings.

Adding  to every effective rule is the correct approach here because CODEOWNERS uses last-match-wins semantics, so a single fallback entry would not preserve the existing path-specific owners. I also verified that  is a valid GitHub user with admin access on this repository, so the review requests should resolve as intended.

The current  failure appears unrelated to this diff; the failing run reports an existing undeclared runtime env ().

slkzgm (commented):
Reviewed the CODEOWNERS change end-to-end. No implementation findings.

Adding `@ElizaReview` to every effective rule is the correct approach here because CODEOWNERS uses last-match-wins semantics, so a single fallback entry would not preserve the existing path-specific owners. I also verified that `ElizaReview` is a valid GitHub user with admin access on this repository, so the review requests should resolve as intended.

The current `env:audit:check` failure appears unrelated to this diff; the failing run reports an existing undeclared runtime env (`WEBHOOK_URL`).

PR #1363, opened 2026-03-26
feat: move agent team widgets into right rail

SYMBaiEX:
## Summary
- Moves agent team widgets (portfolio chart, team stats, recent activity) from the bottom panel into the right sidebar rail
- Simplifies the bottom panel to focus on chat/terminal content
- Consolidates layout so key agent metrics are visible without scrolling

## Test plan
- [ ] Verify agent team page renders with widgets in right sidebar
- [ ] Confirm bottom panel shows chat/terminal content only
- [ ] Check responsive behavior at various viewport widths
- [ ] Verify no regressions in widget data display (portfolio chart, stats, activity)

### DIFF
--- a/apps/web/src/app/agents/team/BottomPanel.tsx
+++ b/apps/web/src/app/agents/team/BottomPanel.tsx
@@ -1,8 +1,7 @@
 'use client';
 
 import { cn } from '@babylon/shared';
-import { Check, ChevronDown, ChevronsUpDown, ChevronUp } from 'lucide-react';
-import { useCallback, useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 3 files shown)


SLKReview (approved):
The layout simplification here is the right direction and the widget rail integration is consistent with the rest of the team page after the follow-up fixes now on this branch.

I pushed a small review commit to close the two concrete issues I found:
- gate `useTeamTradingSummary` behind `rightSidebarOpen` so team summary fetches stop when the rail is hidden
- fall back to `defaultContent` when the active auxiliary tab has already been removed, instead of briefly rendering a blank panel

With those addressed, I do not see remaining blockers in this change set.

PR #1367, opened 2026-03-27
feat(web): redesign agents team page with card view and unified right sidebar

jasperdisney:
## Summary

Redesigns the agents team page with a new card-based agent view and consolidates the bottom panel into the right sidebar for a cleaner, more unified layout.

### Agents Team Page Layout
- **Card/list toggle** for the agent sidebar, with **cards as default** view
- Agent cards display live stats: PnL, trade count, win rate, open positions, and activity timestamps
- Stats fetched from `/api/agents` and `/api/markets/positions` per agent
- **Bottom panel removed** — info tabs (Activity, Wallet, PnL, Logs) now live in the right sidebar
- Right sidebar opens by default; mobile defaults to agents tab
- Entity selector dropdown removed from sidebar header (no longer needed)
- Support `?create=true` query param to open create-agent modal (e.g. from game guide)
- Reply-to-message and conversation delete support added to team chat

### Right Sidebar
- Moved info tab and entity type definitions from the removed BottomPanel component
- Persistent non-closeable info tabs coexist with dynamic closeable message tag tabs
- Removed `entitySelector` prop and `bottomPanelHeight` prop

### Member List
- New card view mode rendering agent stat cards with status indicators
- List view preserved and togglable via header buttons

### Agent Wallet
- Redesigned balance summary from a cramped 3-column horizontal layout to **stacked rows** (label left, value right) — much better readability in narrow sidebar
- Uses `formatCompactCurrency` for consistent number formatting
- Compacted transfer controls and transaction history
- Fixed double `p-4` padding on agent wallet tab content

### Agent PnL
- Compacted layout with collapsible position sections for sidebar width
- Added position count badges and improved empty states

### Agent Logs
- Compacted filter controls and log entry rows for sidebar width
- Smaller typography and tighter spacing

### Agent Edit Modal
- Flattened from multi-step wizard to single-page modal
- AgentRegistry embedded inline instead of lazy-loaded separately

### Agent Registry
- Added `StatusBadge` component for registration state display
- Compacted layout for embedding inside edit modal

### Minor / Auto-fixes
- **TeamPnL**: Minor layout adjustments for sidebar context
- **Tutorial steps**: Updated targets for new sidebar structure
- **Feed pipelines**: Import ordering fix (Biome)
- **MobileHeader**: Tailwind class ordering fix (Biome)

### DIFF
--- a/apps/web/src/app/agents/team/AgentPnL.tsx
+++ b/apps/web/src/app/agents/team/AgentPnL.tsx
@@ -216,66 +216,69 @@ function UserPnL({
         </div>
       </div>
 
-      {/* P&L Summary + Quick Stats */}
-      <div className="grid grid-cols-2 gap-3">
-        {/* P&L Summary (matches profile page) */}
-        <div className="space-y-2 rounded-lg border border-border bg-card/50 p-3">
-          <div className="flex items-center justify-between text-xs">
-            <span className="text-muted-foreground">Total P&L</span>
-            <span
-              className={cn(
-                'font-semibold',
-                isProfitable ? 'text-green-600' : 'text-red-600'
-              )}
-            >
-              {totalPnL >= 0 ? '+' : ''}
-              {formatCompactCurrency(totalPnL)}
-            </span>
-          </div>
-          <div className="flex items-center justify-between text-xs">
-            <span className="text-muted-foreground">Total Assets</span>
-            <span className="font-medium">
-              {formatCompactCurrency(portfolio?.totalAssets ?? 0)}
-            </span>
-          </div>
-          <div className="flex items-center justify-between text-xs">
-            <span className="text-muted-foreground">Available</span>
-            <span className="font-medium">
-              {formatCompactCurrency(portfolio?.available ?? 0)}
-            </span>
-          </div>
-          <div className="border-border border-t pt-2">
-            <div className="flex items-center justify-between text-xs">
-              <span className="text-muted-foreground">In Positions</span>
-              <span className="font-medium">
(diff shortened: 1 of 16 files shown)


SLKReview (changes requested):
I’m requesting changes for two regressions introduced by the new card view.

1. `apps/web/src/app/agents/team/MemberList.tsx`: each card click now calls both `onSelectAgent` and `onTagAgent`. On mobile, `page.tsx` wires those callbacks to conflicting navigation updates (`panel` first, then `chat`), so tapping a card drops the user back into chat instead of opening the new panel view. That breaks the primary “inspect this agent from the card” flow described in the PR.

2. `apps/web/src/app/agents/team/MemberList.tsx`: the card root is a `<button>` that contains another `<button>` for settings and a profile link. Nested interactive controls are invalid HTML and tend to produce React warnings plus unreliable keyboard/screen-reader behavior. The card shell needs to be split from its secondary actions before this ships.

Non-blocking:
- `apps/web/src/app/agents/team/page.tsx` currently fetches open positions sequentially for every agent in the stats effect. That creates an avoidable client-side waterfall on initial render; please parallelize or batch it if we keep this approach.
- I didn’t find automated coverage for the new card interactions. A small test around card selection/navigation would make this refactor safer.

SLKReview (approved):
I re-reviewed the follow-up commit and the blockers from my previous review are addressed.

- Card clicks now only drive agent selection, so the mobile panel flow is no longer fighting with the chat navigation.
- The invalid nested interactive markup in the card view is gone.
- The open-position fetches are now parallelized instead of running as a client-side waterfall.

I did not find any new blocking issues in the delta.

PR #1368, opened 2026-03-27
feat(web): integrate backend logic for agents team redesign [requires #1367 first]

slkzgm:
## Summary
This PR completes the backend/frontend integration layer for the agents team redesign built in #1367.

It introduces a dedicated BFF-style data layer for the redesigned team experience, removes the main client-side waterfalls, and reconnects the redesigned screens to cleaner backend contracts.

## Dependency
This PR is based on the work from #1367 and must be merged after #1367.

Required order:
1. Merge #1367
2. Merge this PR

## What Changed
- Added a team dashboard API tailored to the redesigned agents team page
- Added an agent summary API for sidebar PnL/details loading
- Centralized owned agent summaries and team trading summary aggregation on the server
- Reused a shared user positions snapshot helper behind the existing positions route
- Rewired the redesigned agents team page to the new dashboard integration path
- Rewired agent PnL to consume a single backend snapshot instead of stitching multiple client fetches
- Rewired wallet/team summary consumers to the same server-side source of truth
- Hooked refresh flows after create, edit, and wallet transfer mutations

## Why
The redesign in #1367 introduced the new UI structure, but key data flows were still composed in the client with duplicated logic and request waterfalls.

This PR moves the important aggregation logic to the backend so the redesigned experience is cleaner, more efficient, and easier to maintain.

## Testing
- `bun test apps/web/src/app/api/agents/route.test.ts apps/web/src/app/api/agents/team-dashboard/route.test.ts apps/web/src/hooks/__tests__/useTeamTradingSummary.test.ts`

## Notes
There are still unrelated pre-existing typecheck issues elsewhere in the repository that are outside the scope of this PR.


SLKReview (changes requested):
I pushed `f6c133800` to restore server-side filtering for `/api/agents?autonomousTrading=...`, so that regression is already addressed on the branch.

Remaining blocker:

`apps/web/src/lib/agents/owned-agent-summaries.ts:49`, `apps/web/src/lib/agents/team-dashboard.ts:39`, `apps/web/src/app/api/agents/route.test.ts:95`

The new shared aggregation path now fails fast on any `getPerformance()` / `getAgentConfig()` rejection, and the updated test explicitly locks that behavior in. Before this PR, `/api/agents` degraded per agent and kept the rest of the list usable. With the shared helper, one bad enrichment call now 500s both `/api/agents` and `/api/agents/team-dashboard`, which also takes down the redesigned team page and wallet P&L.

Because the underlying agent rows still exist and these enrichments are auxiliary data, this looks like a reliability regression rather than an improvement. Unless there is a concrete product requirement to hard-fail the entire dashboard on partial enrichment errors, I think the shared helper should preserve the previous best-effort behavior at the per-agent boundary, or otherwise isolate the failure to the affected agent.

SLKReview (dismissed):
Reviewed the follow-up changes through `44b79257a527249b44bb3a9fdc884431e9414747`.

The previous blocker is addressed: `listOwnedAgentSummaries()` now degrades per agent again instead of taking down the whole dashboard on a single enrichment failure, and the added tests cover both the best-effort aggregation path and the new agent summary route.

I don't have further blocking findings on this PR.

PR #1369, opened 2026-03-27
fix(web): align agent click behavior between list and card views

jasperdisney:
## Summary

- **List view** now also calls `onSelectAgent` on click, so the right sidebar updates to show the clicked agent's info (Activity, Wallet, PnL, Logs)
- **Card view** now also calls `onTagAgent` on click, so the agent is tagged in the chat input
- Both views now have identical click behavior: tag agent in chat input + update right panel to that agent's info
- On mobile, `onSelectAgent` runs after `onTagAgent`, so `setMobileView('panel')` correctly wins over `setMobileView('chat')` via React 18 batching

## Files changed

- `apps/web/src/app/agents/team/MemberList.tsx` — unified click handlers for both list and card view modes

### DIFF
--- a/apps/web/src/app/agents/team/MemberList.tsx
+++ b/apps/web/src/app/agents/team/MemberList.tsx
@@ -312,12 +312,11 @@ export function MemberList({
                   onTagAgent?.(agent);
                   onSelectAgent?.(agent.id);
                 }}
-                disabled={isProcessing}
                 className={cn(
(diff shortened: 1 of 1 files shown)


SLKReview (approved):
Pushed a small follow-up commit (`ee2ec4880`) to keep the list and card behaviors aligned when an agent is processing, and to update the list action label to match the combined behavior.

With that in place, I did not find any remaining blocking issues in this PR.

PR #1388, opened 2026-03-30
refactor: DI cleanup for QuestionManager, GameLoop + fix ghost variables

revlentless:
## Summary

- **Fix renderPrompt ghost variables**: Unpopulated optional template vars (e.g. `{{characterRoster}}`, `{{previousTrades}}`) were being sent as literal text to the LLM. Now replaced with empty string.
- **QuestionManager DI**: Constructor accepts optional `services` object for `MarketContextService`, `MarketDecisionEngine`, `TradeExecutionService`. Falls back to `new` when not provided — fully backward compatible.
- **GameLoop DI**: Constructor accepts optional 5th arg for `TradeExecutionService` and `PerpMarketService`. Same fallback pattern.

## Why

Prerequisite for the dev tools suite (context-inspector, question-gen-bench, arc-simulator, trading-replay). These components couldn't be tested in isolation because they hardcoded `new` for their dependencies. The ghost variable fix is also a standalone production bug — 6 template placeholders were being sent verbatim to the LLM in every NPC trading decision prompt.

## Test plan

- [ ] `bun run check` passes
- [ ] Existing callers of `QuestionManager(llm)` and `GameLoop(world, feed, decisions, relationships)` still work (no signature break)
- [ ] Verify rendered trading prompts no longer contain literal `{{variableName}}` strings

### DIFF
--- a/packages/engine/src/GameLoop.ts
+++ b/packages/engine/src/GameLoop.ts
@@ -56,25 +56,24 @@ export interface SimulationTickResult {
  *
  * Used by both live game ticks (cron jobs) and game simulation (full game generation).
  */
+export interface GameLoopServices {
+  tradeExecutionService?: TradeExecutionService;
+  perpMarketService?: PerpMarketService;
+}
+
 export class GameLoop {
   private trendingTopics?: TrendingTopicsEngine;
   private recentPosts: FeedPost[] = [];
   private tickCount = 0;
 
-  private injectedServices?: {
-    tradeExecutionService?: TradeExecutionService;
-    perpMarketService?: PerpMarketService;
-  };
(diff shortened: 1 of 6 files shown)


SLKReview (dismissed):
This looks good to me.

The `renderPrompt` change is minimal and correctly scoped to the production issue described in the PR, and the DI changes in `QuestionManager` and `GameLoop` stay narrow and backward-compatible. I did not find dead branches, duplicated logic, or unnecessary compatibility layers beyond the intentional optional-service fallback for tooling.

One follow-up I would still recommend separately is adding a small unit test around optional placeholder cleanup in `renderPrompt`, since that is the production-facing bug being fixed here.

PR #1390, opened 2026-03-30
feat: market-diversity-report dev tool

revlentless:
## Summary

- New CLI script `scripts/market-diversity-report.ts` that audits the active prediction market pool for diversity
- Computes: topic clustering (Jaccard similarity), entity frequency, timeframe balance, category distribution, daily topic concentration, near-duplicate detection
- Run: `bun run report:markets`

## Why

From the stories/markets analysis: all markets cluster around a single daily topic, the same actors/orgs dominate every question, and near-duplicate questions slip through prompt-only dedup. This tool gives us a diversity baseline to measure against as we implement multi-topic support and structural dedup.

## Usage

```bash
bun run report:markets                    # current snapshot
bun run report:markets -- --verbose       # include full question texts
bun run report:markets -- --history 7     # trend over last 7 days
```

## Test plan

- [ ] `bun run report:markets` runs against dev DB
- [ ] Handles empty DB gracefully

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,8 @@
     "privy:setup:solana-policy": "bun run scripts/create-privy-solana-policy.ts",
     "solana:registration:debug": "bun run scripts/debug-solana-registration.ts",
     "seed:nft-snapshot:csv": "bun run scripts/seed-nft-snapshot-csv.ts",
-    "users:export:newsletter": "bun run scripts/export-newsletter-users-csv.ts"
+    "users:export:newsletter": "bun run scripts/export-newsletter-users-csv.ts",
+    "report:markets": "bun run scripts/market-diversity-report.ts"
   },
   "dependencies": {
     "@a2a-js/sdk": "^0.3.5",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "solana:registration:debug": "bun run scripts/debug-solana-registration.ts",
     "seed:nft-snapshot:csv": "bun run scripts/seed-nft-snapshot-csv.ts",
     "users:export:newsletter": "bun run scripts/export-newsletter-users-csv.ts",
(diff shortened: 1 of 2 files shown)


SLKReview (dismissed):
I pulled the branch locally and pushed a follow-up commit (`7d721b492`) before submitting this review.

That fix addressed the two blocking issues I found:
- the topic clustering logic now merges connected components correctly instead of leaving overlapping clusters split across sets
- the report now uses the raw Drizzle client for its read-only queries, so the command is executable in a normal local dev setup without depending on the broken lazy top-level `db.select()` proxy path

Verification:
- `bunx biome check package.json scripts/market-diversity-report.ts`
- `bun --env-file=/home/slk/work/babylon/monorepo/.env run scripts/market-diversity-report.ts --history 1` now reaches the DB query path correctly; execution stops later because the local Postgres on `localhost:5433` is not running in this environment
- repo-wide typecheck still fails on pre-existing `NotificationData` typing errors in `packages/api/src/services/achievement-service.ts`, unrelated to this PR

SYMBaiEX (approved):
## Approved ✅

The blocking JSON syntax issue (missing comma in `package.json`) is fixed and CI can parse the file again. Review feedback addressed:

| Issue | Status |
|---|---|
| Missing comma in `package.json` (CI blocker) | Fixed in `e6c1619` |
| `--history` input validation (NaN silently ignored) | Fixed — now errors on non-numeric input |
| DB connection cleanup (`process.exit` hygiene) | Fixed — `.finally()` ensures cleanup on all paths |
| O(n²) pairwise comparisons | Acceptable for expected market counts (<100); noted for future |
| Full org table fetch | Fine for read-only report at current scale |
| ANSI in redirected output | Nice-to-have, not blocking |

Script is clean, follows repo conventions, and fills a real diagnostic need. Good to merge.

PR #1391, opened 2026-03-30
feat: prompt-diff dev tool

revlentless:
## Summary

- New CLI script `scripts/prompt-diff.ts` that compares two prompt template versions rendered with the same context
- Shows metadata diff (temperature, maxTokens), per-section token comparison, and unified text diff with color
- Supports git refs (`git:HEAD~1:path/to/prompt.ts`) for comparing against previous versions

## Why

Most fixes for the stories/markets/agent-context issues involve prompt changes. Without this tool, evaluating a prompt change means deploying and hoping. With it, you can see exactly what changed in the rendered output, section by section, with token impact.

## Usage

```bash
# Compare current vs previous version of a prompt
scripts/prompt-diff.ts --old git:HEAD~1:packages/engine/src/prompts/trading/npc-market-decisions.ts \
                       --new packages/engine/src/prompts/trading/npc-market-decisions.ts

# Compare two prompt files with context
scripts/prompt-diff.ts --old prompts/v1.ts --new prompts/v2.ts --context snapshot.json

# Section-level summary only
scripts/prompt-diff.ts --old prompts/v1.ts --new prompts/v2.ts --section-only
```

## Test plan

- [ ] Script runs without DB connection (no DB dependency)
- [ ] Git ref loading works (`git:HEAD~1:...` syntax)


SLKReview (changes requested):
Two correctness issues need to be fixed before this can land.

1. `scripts/prompt-diff.ts:181-196` breaks the advertised `git:<ref>:<path>` flow for real prompts that import anything outside `packages/engine/src/prompts`. The rewrite currently sends every `../` import to `packages/engine/src/prompts/...`, so `../../utils/randomization` becomes `packages/engine/src/prompts/utils/randomization`, which does not exist. I verified this with:

`bun scripts/prompt-diff.ts --old git:HEAD:packages/engine/src/prompts/trading/npc-market-decisions.ts --new packages/engine/src/prompts/trading/npc-market-decisions.ts --section-only`

and it fails with `Cannot find module .../packages/engine/src/prompts/utils/randomization`. Since git-based comparison is one of the headline features in the PR description, this is a functional blocker.

2. `scripts/prompt-diff.ts:116-126` reimplements template rendering instead of reusing the existing prompt utility, and it changes behavior by stripping every unresolved `{{placeholder}}`. The engine's `renderTemplate` in `packages/engine/src/prompts/define-prompt.ts:77-86` does not do that. In practice this makes the diff output misleading when the context is partial or empty: missing variables disappear from the rendered prompt, section token counts collapse, and prompt changes caused by placeholder renames can be hidden entirely. The tool should either reuse the existing renderer or match its semantics exactly.

Once those are fixed, the rest of the CLI structure looks reasonable.

SYMBaiEX (approved):
✅ Approved — nice ergonomic addition to the devtools suite.

**Issues found & fixed (pushed):**

1. **Import rewriting breaks for imports outside `prompts/`** — the original regex hardcoded `packages/engine/src/prompts/` as the resolution base, so `../../utils/randomization` (which goes up past `prompts/`) resolved to a wrong path. Fixed to use `path.resolve` from the original file's directory in the repo.

2. **`renderTemplate` strips unresolved `{{vars}}`** — the engine's `renderTemplate` preserves unresolved placeholders, but the script was deleting them. This hid placeholder renames in diffs and made token counts misleading. Fixed by removing the stripping line.

3. **`diff` not declared as a direct dependency** — `diff` and `@types/diff` were only transitive. Added both as explicit devDependencies.

Commit: `fix(review): resolve import rewriting, preserve unresolved template vars, add diff dep`

PR #1392, opened 2026-03-30
feat: enrich agent trading context — populate ghost variables, fix truncation, add memories

revlentless:
## Summary

Addresses recommendations from `docs/agent-context-analysis.md`. Agents were making trading decisions with severely impoverished context — 6 template variables were empty (literal `{{varName}}` sent to the LLM), posts were truncated to 200 chars, questions to 120 chars, and group chats showed only 2 messages at 120 chars.

### Tier 1: Wire up what already exists
- **Populate `{{resolvedQuestionsContext}}`** — last 10 resolved questions with YES/NO outcomes, so agents can learn from past market resolutions
- **Populate `{{previousTrades}}`** — last 24h of NPC trades with action, price, and reasoning, so agents see what they and others have done
- **Append NPC memories to trader dashboards** — `NpcMemoryService.formatMemoriesForPrompt()` already exists for posting; now wired into trading
- **Fix `renderPrompt` ghost variables** — unpopulated optional vars now replaced with empty string instead of sent as literal template text
- **Upgrade reality grounding** from `'minimal'` to `'concise'` for richer world context

### Tier 2: Fix truncation strategy
| Data | Before | After |
|------|--------|-------|
| Feed posts | 50 posts × 200 chars | 15 posts × 500 chars |
| World events | 30 events × 150 chars | 20 events × 300 chars |
| Prediction questions | Truncated at 120 chars | **Never truncated** |
| Group chat messages | 15 msgs × 120 chars | 8 msgs × 300 chars |
| Dashboard positions | Top 3 by PnL | **All positions** |
| Dashboard group chat | 2 messages shown | 5 messages shown |
| Dashboard relationships | Max 4 | Max 6 |
| "Current Focus" | 3 posts × 20 chars | **Removed** (was noise) |

### Tier 3: Extended capabilities
- **Autonomous agent group chat lookback**: 24 hours / 15 messages (was 1 hour / 10 messages)

## Test plan
- [ ] `bun run typecheck` passes (pre-existing `achievement-service.ts` errors are unrelated)
- [ ] `bun run check` passes
- [ ] Deploy to staging and verify NPC trading decisions reference resolved questions and past trades
- [ ] Verify rendered trading prompts no longer contain literal `{{variableName}}` strings
- [ ] Verify prediction market questions appear untruncated in market table

### DIFF
--- a/packages/agents/src/autonomous/AutonomousGroupChatService.ts
+++ b/packages/agents/src/autonomous/AutonomousGroupChatService.ts
@@ -76,16 +76,16 @@ export class AutonomousGroupChatService {
         continue;
       }
 
-      // Get recent messages in this group
-      const oneHourAgo = new Date(Date.now() - 60 * 60 * 1000);
+      // Get recent messages in this group (24h lookback for strategic continuity)
+      const oneDayAgo = new Date(Date.now() - 24 * 60 * 60 * 1000);
       const recentMessages = await db
         .select()
         .from(messages)
         .where(
-          and(eq(messages.chatId, chat.id), gte(messages.createdAt, oneHourAgo))
+          and(eq(messages.chatId, chat.id), gte(messages.createdAt, oneDayAgo))
         )
         .orderBy(desc(messages.createdAt))
-        .limit(10);
+        .limit(15);
 
       if (recentMessages.length === 0) continue;
 
--- a/packages/engine/src/MarketDecisionEngine.ts
+++ b/packages/engine/src/MarketDecisionEngine.ts
@@ -67,6 +67,7 @@
  */
 
 import {
+  actorState,
   and,
   db,
   desc,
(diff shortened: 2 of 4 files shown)


SLKReview (changes requested):
I still see two blocking issues before this is ready to merge.

1. `packages/engine/src/MarketDecisionEngine.ts:2286`
   `getCachedResolvedQuestions()` is formatting the resolved history from `q.outcome`, but in this codebase the final resolution is stored in `resolvedOutcome` when a question is resolved. `outcome` is the originally planned/expected answer, so any market that resolves opposite to the planned outcome will teach the trading prompt the wrong lesson. This needs to read `resolvedOutcome` (and guard/skip nulls) instead.

2. `packages/engine/src/MarketDecisionEngine.ts:2333`
   `getMemoriesForNPCs()` now does one `NpcMemoryService.getRecentMemories()` query per NPC. On a normal 20-NPC batch that adds 20 extra DB round trips to every tick, even though these memories live on `actorState.recentMemories` and the codebase already has batched `IN (...)` reads for actor state. I don't think we should merge the context enrichment with that avoidable latency regression; this should be batched into a single read / helper first.

revlentless:
Superseded by #1401 which combines this PR with #1394, #1399, and new work (item #16).

PR #1395, opened 2026-03-30
feat: add profile-driven perp market simulation

slkzgm:
## Summary
- replace the uniform perp volatility simulator with profile-driven market dynamics
- introduce deterministic per-market simulation profiles with different volatility, jump, beta, liquidity, and trend characteristics
- add a lightweight global market regime state so all perps share a common drift/volatility backdrop while retaining idiosyncratic behavior
- make open-interest dampen simulated noise so deeper markets move less mechanically than thin ones
- cover the new simulation layer with focused unit tests

## Why
This is the next realism step after the market-context coherence pass. The existing simulator made most perp markets behave too similarly because they all used the same generic noise process. This change gives each market a stable identity and a latent fair-value/regime process without yet changing execution or introducing full microstructure.

## Testing
- `bun test packages/engine/src/services/market-simulation-profiles.test.ts`
- `bunx tsc -p packages/engine/tsconfig.json --noEmit`

## Follow-ups
- synthetic spread/depth and nonlinear impact
- better separation of fair value vs tradeable/quoted price
- richer calibration per asset class and event sensitivity

### DIFF
--- a/packages/engine/src/game-tick.ts
+++ b/packages/engine/src/game-tick.ts
@@ -83,6 +83,14 @@ import {
   WalletService,
   worldFactsGenerator,
 } from './services';
+import {
+  buildMarketSimulationProfile,
+  createInitialMarketSimulationState,
+  evolveGlobalMarketSimulationState,
+  type GlobalMarketSimulationState,
+  generateProfileDrivenMarketMove,
+  getDefaultGlobalMarketSimulationState,
+} from './services/market-simulation-profiles';
 // Note: ActorSocialActions, FollowingMechanics, processNPCSocialEngagements,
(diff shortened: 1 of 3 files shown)


SLKReview (approved):
No remaining blocking findings.

I pushed `ff2bee63e` while reviewing to fix one functional gap in the new simulation layer: `latentPrice` was being stored in `MarketSimulationState` but never read back, which meant the latent fair-value process reset to `currentPrice` every tick instead of persisting across ticks. The follow-up commit now evolves fair value from the prior latent state and adds a regression test covering that behavior.

Verified with:
- `bun test packages/engine/src/services/market-simulation-profiles.test.ts`

PR #1396, opened 2026-03-30
feat: add synthetic perp execution microstructure

slkzgm:
## Summary
- add a synthetic perp microstructure layer with bid/ask quotes, spread, depth, and nonlinear slippage
- execute perp opens, closes, adds, and flips against synthetic quotes instead of the raw market price
- expose the microstructure helper via `@babylon/core/markets/perps`
- update perp service tests to validate economic behavior under execution friction
- add focused unit tests for quote generation and thin-vs-deep market behavior

## Why
Phase 2 gave each perp a more realistic latent/fair-value process, but execution was still too frictionless. This change is the next step: trades now experience synthetic market microstructure, so thin markets slip more, spreads matter, and execution feels less like filling against a single abstract price.

## Testing
- `bun test packages/core/markets/perps/__tests__/microstructure.test.ts`
- `bunx tsc -p packages/core/tsconfig.json --noEmit`

## Follow-ups
- quote resilience / recovery over time
- exposing bid/ask/spread/depth through API/UI
- converging trade-impact and quote-state evolution into a single canonical market-state model

### DIFF
--- a/packages/core/markets/perps/PerpMarketService.ts
+++ b/packages/core/markets/perps/PerpMarketService.ts
@@ -1,4 +1,5 @@
 import { logger, PERP_MARKET_CONFIG } from '@babylon/shared';
+import { getSyntheticPerpExecutionPrice } from './microstructure';
 import type {
   PerpCloseInput,
   PerpDbPort,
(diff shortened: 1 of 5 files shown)


SLKReview (approved):
I reviewed this locally and pushed one small hardening fix on top: `c29404510`.

The main issue I found was in the new exported microstructure helper: it only validated `currentPrice`, so `NaN` / non-finite `markPrice` or `indexPrice` values could still leak through the fallback chain and produce non-finite quote fields for downstream consumers. The follow-up commit now normalizes those fallback references before quote construction and adds a focused regression test for that case.

With that fix in place, the feature looks good to me. The service integration is consistent with the current perp execution model, the microstructure helper stays small and readable, and the new tests cover the key economic behaviors plus the invalid-input edge case.

Validated locally with:
- `bun test packages/core/markets/perps/__tests__/microstructure.test.ts`

revlentless:
LGTM!

PR #1405, opened 2026-03-31
fix: remove incorrect * 100 price scaling on prediction market positions

revlentless:
## Summary

Fixes the root cause of agent bankruptcy and corrupted PnL data (items S7/S8 from the agent context audit).

### The Bug

`trade-execution-service.ts` multiplied `avgPrice` by 100 before storing it as `entryPrice` on `PoolPosition`:

```typescript
const entryPrice = result.avgPrice * 100;  // BUG
```

This assumed `avgPrice` is a 0-1 probability that needs converting to cents. But the CPMM's `avgPrice` is actually **cost-per-share** (`usdAmount / sharesBought`), which is typically 0.5-3.0+ depending on trade size and market depth.

Example: A $2000 buy on a 50/50 market with 10k liquidity:
- `sharesBought = 1428.57`
- `avgPrice = 2000 / 1428.57 = 1.4` (cost per share)
- **Before fix**: `entryPrice = 1.4 * 100 = 140` (wrong)
- **After fix**: `entryPrice = 1.4` (correct)

This 100x inflation compounded through settlements, creating the trillion-dollar negative PnL values observed on all autonomous agents.

### Changes

All `* 100` conversions removed from:
- `entryPrice` on buy path
- `currentPrice` on buy + conflict-update paths
- `price` on sell paths (npcTrades recording)
- `executionPrice` on sell return values

### Verified on local DB

Reproduced on a fresh local Postgres. Before the fix, `PoolPosition.entryPrice` ranged from 9.7 to 1631.5 (should be 0.1-3.0). The `* 100` scaling is clearly the cause.

### Note on existing data

Existing production data has corrupted prices. After merging this fix, a data migration script should be run to:
1. Divide all `PoolPosition.entryPrice` values by 100
2. Divide all `PoolPosition.currentPrice` values by 100
3. Recalculate `unrealizedPnL` for open positions

This PR does NOT include the migration to keep the fix focused. The migration should be a separate PR.

## Test plan
- [x] `bun run check` passes
- [x] `bun run test:unit` — 154 pass, 2 pre-existing failures
- [x] 6 new tests verify CPMM avgPrice is cost-per-share (not probability)
- [ ] Run game ticks on clean local DB and verify PoolPosition.entryPrice values are < 5.0
- [ ] Write and run data migration for existing production data


lalalune:
Changes merged into shaw/research (will land on staging via PR #1408)

SLKReview (approved):
Reviewed the remaining delta in this PR.

Removing the last two `* 100` multipliers on the sell/close write paths is the correct fix: `PredictionMarketService` returns `market.yesPrice` / `market.noPrice` as raw 0..1 probabilities, and the buy path already stores those raw values on `staging`. With this change, the prediction position lifecycle is internally consistent again across open, sell, and close flows.

I did not find any blocking issues in correctness, code quality, DRY, or performance for the current diff.

Verified locally:
- `bun test packages/testing/unit/prediction-price-scaling.test.ts --preload ./packages/testing/unit/preload.ts`
- `bunx --bun @biomejs/biome check packages/engine/src/services/trade-execution-service.ts packages/testing/unit/prediction-price-scaling.test.ts`

SYMBaiEX (approved):
LGTM

PR #1406, opened 2026-03-31
feat: data migration script for prediction market price correction

revlentless:
## Summary

Companion to #1405. Provides a migration script to fix existing `PoolPosition` data that was stored with the incorrect `* 100` scaling.

### Usage
```bash
# Dry run (report only, safe to run anytime)
bun run scripts/fix-prediction-prices.ts

# Apply the fix
bun run scripts/fix-prediction-prices.ts --apply
```

### What it does
1. Queries all prediction-type `PoolPosition` records
2. Reports current min/max/avg for `entryPrice` and `currentPrice`
3. Auto-detects if data needs fixing (max entryPrice > 10 indicates scaling)
4. Divides `entryPrice` and `currentPrice` by 100 for all prediction positions
5. Reports after-fix statistics

### Deployment order
1. Merge #1405 first (fixes the code that writes prices)
2. Deploy
3. Run this migration: `bun run scripts/fix-prediction-prices.ts --apply`

## Test plan
- [x] Dry run works correctly with no data
- [x] Dry run correctly detects "no fix needed" when prices are normal
- [ ] Run against staging/production DB in dry-run mode first
- [ ] Apply fix and verify price ranges are reasonable

### DIFF
--- a/scripts/fix-prediction-prices.ts
+++ b/scripts/fix-prediction-prices.ts
@@ -1,0 +1,127 @@
+#!/usr/bin/env bun
+/**
+ * Fix Prediction Market Price Data
+ *
+ * Corrects PoolPosition entries where entryPrice and currentPrice were
+ * stored with an incorrect * 100 scaling. The CPMM's avgPrice is
+ * cost-per-share (typically 0.5-3.0+), not a 0-1 probability, so the
+ * * 100 multiplication created values 100x too large.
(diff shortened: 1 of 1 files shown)


lalalune:
Changes merged into shaw/research (will land on staging via PR #1408). Deploy order: #1405 code first, then run migration script.

PR #1407, opened 2026-03-31
fix: shuffle actors/orgs in production question generation + restore analysis doc

revlentless:
## Summary

Two changes:

### 1. Shuffle actors/orgs in production question generation

Both production question generation paths loaded actors and organizations from `StaticDataRegistry` in deterministic order. The LLM saw the same ordered list every tick, producing repetitive questions about the same characters.

**Before**: `StaticDataRegistry.getAllActors().filter(...).slice(0, 30)` — same 30 actors in same order every time

**After**: `shuffleArray(StaticDataRegistry.getAllActors().filter(...)).slice(0, 30)` — random 30 actors each tick

Applied to both paths:
- `generateQuestionsForContinuousGame()` (main production path)
- The timeframed question generation path

`shuffleArray` was already imported but never used in these paths.

### 2. Restore stories-markets-analysis.md

The analysis doc was lost during branch cleanup. Restored from commit history and updated shuffle status from "CONFIRMED broken" to "FIXED".

## Test plan
- [x] `bun run check` passes
- [x] `bun run test:unit` — 155 pass, 0 failures
- [ ] Run `bun run report:markets` after a few ticks to verify improved entity diversity

### DIFF
--- a/docs/stories-markets-analysis.md
+++ b/docs/stories-markets-analysis.md
@@ -1,0 +1,231 @@
+# Deep Analysis: Stories & Prediction Markets Systems
+
+> Audited against the codebase on 2026-03-30. Each finding is annotated with its verification status.
+
+## Architecture Overview
+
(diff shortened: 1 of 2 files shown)


odilitime (approved):

lalalune:
Changes merged into shaw/research (will land on staging via PR #1408)

PR #1410, opened 2026-03-31
feat: multi-topic market generation — spread markets across 3 themes instead of 1

revlentless:
## Summary

Addresses the #1 repetition driver from `docs/stories-markets-analysis.md`: the single daily topic bottleneck that forces all 10 market slots into one theme.

### The Problem

Every market created in a tick gets the same daily topic. If RSS picks "AI regulation" as the topic, all 10 markets are about AI regulation. When RSS fails, it falls back to yesterday's topic, extending the echo chamber across days.

### The Fix

Instead of fetching 1 topic, fetch 3 candidates and rotate:

```
Market slot 1 (3d): Topic A (AI regulation)
Market slot 2 (2d): Topic B (tech earnings)
Market slot 3 (1d): Topic C (crypto policy)
Market slot 4 (12h): Topic A (rotates back)
Market slot 5 (6h): Topic B
...
```

### Changes

**Daily Topic Service** (`daily-topic-service.ts`):
- `getTopicCandidatesForDate(date, count=3)` — returns primary topic + additional RSS candidates
- `isTextOnAnyTopic(text, topics[])` — filter against multiple topics
- `buildMultiTopicPromptContext(topics[])` — LLM instruction for multi-topic generation

**Markets Tick** (`markets-tick/route.ts`):
- Fetch 3 topic candidates instead of 1
- Rotate `topicRotationIndex` per market creation call
- Both replacement markets and gap-fill markets use rotation

### Design Decisions
- **No schema changes** — the `dailyTopics` table UNIQUE on date stays. Multi-topic rotation happens at market creation, not storage
- **Graceful degradation** — if RSS only produces 1 candidate, all markets get that topic (same as before). If 0, markets are skipped (same as before)
- **Backward compatible** — `ensureTopicForDate()` still works for code that needs a single primary topic

## Test plan
- [x] `bun run check` passes
- [x] `bun run test:unit` — 155 pass, 0 failures
- [ ] Run `bun run report:markets` after several ticks to verify topic diversity across markets
- [ ] Verify questions span multiple topics instead of clustering on one


SLKReview (changes requested):
Two issues block this for me before merge:

1. `packages/engine/src/services/daily-topic-service.ts:527` breaks the existing manual override contract. `ensureTopicForDate()` correctly returns the locked/manual topic for the day, but `getTopicCandidatesForDate()` immediately appends fresh auto candidates and `markets-tick` rotates across them. If an admin locks the daily topic, only one market keeps that topic and the rest can drift to RSS-derived topics. That defeats the purpose of `setManualTopic(..., isLocked: true)`. This method should short-circuit to `[primary]` when `primary.isLocked` (or otherwise preserve the lock semantics).

2. `packages/engine/src/services/daily-topic-service.ts:299` and `packages/engine/src/services/daily-topic-service.ts:312` add exported helpers that are never used anywhere in the new flow. The implementation still drives `QuestionManager.generateTimeframeQuestion()` with a single `DailyTopicContext` per market, so `isTextOnAnyTopic()` and `buildMultiTopicPromptContext()` are speculative surface area / dead code right now. I would remove them until a caller actually needs multi-topic prompt/filtering.

Once those are addressed, the topic rotation in `markets-tick` looks directionally right.

PR #1411, opened 2026-03-31
feat: generate agent avatars with fal and idempotent API

odilitime:
Default agent profile images use /assets/agent-monkeys/ presets with shared helpers. New POST /api/agents/generate-avatar runs fal (nano-banana-2 edit when reference is available), uploads to profile storage, and dedupes via Idempotency-Key plus cache and singleflight. Create flow uses session idempotency, ref-guarded template apply, and aborts only the avatar fetch. Adds reference.webp, AGENT_AVATAR_REFERENCE_IMAGE_URL in .env.example, and GENERATE_AGENT_AVATAR rate limit.

Made-with: Cursor


SLKReview (changes requested):
Requesting changes for three blocking issues:

1. `packages/shared/src/utils/assets.ts` now makes every default agent avatar point to `/assets/agent-monkeys/monkey-*.jpg`, and the create/edit surfaces have all been switched over to that path. In this branch the repository only contains `apps/web/public/assets/agent-monkeys/reference.webp`; there are no `monkey-*.jpg` assets to serve. That means the new default avatar UX 404s locally and in preview unless there is an out-of-band asset deployment that is neither part of nor documented by this PR.

2. `apps/web/src/app/agents/create/hooks/useAgentForm.ts` stores the avatar idempotency key in `sessionStorage`, but only clears it inside `clearDraft()`, which is called after a successful create. Closing or backing out of the flow does not reset that key, so reopening agent creation in the same tab reuses the cached avatar for up to 15 minutes instead of generating a new one. That is broader than “one create session” and makes repeated create attempts unexpectedly sticky.

3. `loadTemplate()` in `apps/web/src/app/agents/create/hooks/useAgentForm.ts` rethrows non-abort errors from an async function that is invoked inside `useEffect` and never awaited. Any network/JSON/auth failure on that path becomes an unhandled promise rejection in the browser. Please catch and log inside the effect (and leave the form initialized) rather than throwing out of the effect body.

After those are addressed, I would also add targeted tests around the new avatar idempotency helper/route, since the singleflight + cache behavior is subtle and currently untested.

PR #1412, opened 2026-03-31
feat: structural question deduplication — 3-layer similarity checking

revlentless:
## Summary

Adds programmatic similarity checking to prevent near-duplicate prediction market questions. The LLM is told "don't repeat" via prompt instructions, but it frequently generates variations of the same question ("Will TeslAI deploy to cause a dill dilemma?" vs "Will TeslAI deploy to stage a fennel feud?"). This PR adds structural enforcement.

### Three-layer detection

**Layer 1: Token Jaccard (threshold: 45%)**
Extracts significant tokens (after stop word removal), computes Jaccard similarity. Catches obvious rewording.

**Layer 2: Entity overlap + Jaccard (threshold: 50% entity + 40% token)**
Extracts known actor/org names via StaticDataRegistry. If two questions share the same subjects AND have moderate token overlap, they're flagged. Calibrated to allow same-entity-different-predicate (e.g., "Will X announce Y?" vs "Will X resign?").

**Layer 3: Template detection**
Replaces entity names with `[ENTITY]`, numbers with `[NUM]`, days with `[DAY]`. If two questions produce identical templates, one is rejected. Catches "Will [ENTITY] announce partnership with [ENTITY] by [DAY]?" patterns.

### Integration point

Inserted in `generateQuestionsForContinuousGame()` after text sanitization, before DB insert. Checks each new question against:
- All active questions (up to 20)
- Recently resolved questions (last 7 days, up to 10)
- Previously accepted questions in the same batch (first-wins)

Rejected questions are logged and skipped. No retry — the LLM already generated them, we just filter.

### Test coverage

21 tests covering all layers, edge cases, and batch behavior.

## Test plan
- [x] `bun run check` passes
- [x] `bun run test:unit` — 156 pass, 0 failures
- [x] 21 dedup-specific tests all pass
- [ ] Run game ticks and verify duplicate questions are logged and rejected
- [ ] Run `bun run report:markets` to verify question diversity improves

### DIFF
--- a/packages/engine/src/QuestionManager.ts
+++ b/packages/engine/src/QuestionManager.ts
@@ -100,6 +100,7 @@ import { MarketMetricsService } from './services/market-metrics-service';
 import { saveArcPlan } from './services/narrative-state-service';
 import { ensureMarketOnChain } from './services/onchain-market-service';
 import { QuestionArcPlanner } from './services/question-arc-planner';
+import { checkQuestionSimilarity } from './services/question-dedup-service';
 import { StaticDataRegistry } from './services/static-data-registry';
 import { TradeExecutionService } from './services/trade-execution-service';
 import type {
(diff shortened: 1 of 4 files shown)


odilitime:
does it not cancel out everything? how many retries probably should be tracked, so we don't have this burn llm costs

revlentless:
@odilitime 
Good concerns. Two answers:

**Does it cancel everything?** No — the thresholds are calibrated to only catch near-duplicates. Genuinely different questions pass easily. The test suite proves this: "Will the SEC classify ZcAIsh?" vs "Will NVIDAI release a new GPU?" scores 0% similarity and passes. Only questions with >45% token overlap, or shared entities + >40% overlap, or identical templates (same scaffold with swapped names) get rejected.

**LLM cost burn:** The dedup is a post-generation filter, not a retry loop. Zero extra LLM calls. The LLM generates N questions, we filter duplicates, whatever survives gets created. If the filter rejects 2 out of 5, we still create 3 — no regeneration. But you're right that we should track rejection rates. I'll add logging that tracks accepted/rejected counts per tick so we can monitor if the LLM is consistently wasting generations.

PR #1413, opened 2026-03-31
fix(ci): add CI_ADMIN_TOKEN for integration test admin auth

revlentless:
## Summary
- Admin integration tests (52 of 88 failures) fail because the server runs with `NODE_ENV=production`, which disables the dev admin token path
- Add `CI_ADMIN_TOKEN` env var support to admin middleware — only checked when `CI=true`
- Add shared `getAdminToken()` helper to integration test helpers
- Update all 8 admin test files to use the shared helper
- Add integration test step to staging CI workflow (was only on production)

## Root Cause
The `requireAdmin` middleware only accepts `x-dev-admin-token` when `isDevelopment` is true (`NODE_ENV !== 'production'`). In CI, the Next.js server must run with `NODE_ENV=production` for a proper production build test, so the dev token path is disabled.

## Security
- `CI_ADMIN_TOKEN` is only checked when `process.env.CI === 'true'`
- The token is a static value set in the CI workflow, not a secret (it's only used against the ephemeral CI database)
- Production deployments never have `CI=true`

## Test plan
- [ ] Integration tests for admin endpoints pass with `CI_ADMIN_TOKEN`
- [ ] Quality Gate (unit tests + typecheck + lint) passes
- [ ] Existing dev token auth still works locally

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -807,6 +807,9 @@ ALLOW_TEST_PRIVY_DID_AUTH=
 ANTHROPIC_API_KEY=
 ATROPOS_API_URL=
 AVAILABLE_VRAM_GB=
+# Optional: static admin token for CI integration tests. Only checked when CI=true.
+# Set to any string value. The same value must be used by both server and test runner.
+CI_ADMIN_TOKEN=
 BABYLON_A2A_ENDPOINT=
 BABYLON_AGENT0_PRIVATE_KEY=
 BABYLON_DIAMOND_ADDRESS=
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -277,6 +277,29 @@ jobs:
           # NEXT_PUBLIC_* variables are embedded at build time, but ensure they're available
           # at runtime for any server-side rendering or API routes that might need them
           NEXT_PUBLIC_PRIVY_APP_ID: ${{ secrets.NEXT_PUBLIC_PRIVY_APP_ID || secrets.PRIVY_APP_ID || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork && 'cl_ci_fork_dummy') || '' }}
+          CI_ADMIN_TOKEN: ci-integration-test-admin-token
+
+      - name: Run Integration Tests
+        continue-on-error: true
+        timeout-minutes: 10
+        run: |
+          echo "🧪 Running integration tests..."
+          bun test packages/testing/integration/ \
+            --preload ./packages/testing/integration/preload.ts \
+            --timeout 30000
+        env:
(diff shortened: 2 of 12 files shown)


SLKReview (changes requested):
I found two blocking issues before this can be merged.

1. `packages/api/src/admin-middleware.ts:149` returns a synthetic `ci-admin-user` that is not backed by a real `User` row, but several admin write paths persist `admin.userId` / `admin.dbUserId` into FK-backed columns. For example, `apps/web/src/app/api/admin/whitelist/route.ts:77` writes `grantedBy`, `apps/web/src/app/api/admin/whitelist/config/route.ts:76` writes `updatedBy`, and both columns reference `users.id` in `packages/db/src/schema/whitelist.ts:27` and `packages/db/src/schema/whitelist.ts:54`. The same problem exists for role grants in `apps/web/src/app/api/admin/roles/route.ts:169` with the FK declared in `packages/db/src/schema/admin.ts:92`. So this auth bypass is not a general admin identity for CI: it will still 500 as soon as an integration test hits one of those endpoints. The CI path needs to reuse or provision a real persisted admin user instead of returning a fake in-memory identity.

2. `packages/testing/integration/helpers.ts:73` regresses the test helper contract by dropping the existing `DEV_ADMIN_TOKEN` / `TEST_ADMIN_TOKEN` overrides that the individual test files were using before this refactor. After this change, `getAdminToken()` only returns `CI_ADMIN_TOKEN` or a locally-derived dev token, so any local/staging harness that injects an explicit admin token now loses authenticated coverage. Since the point of the helper is to centralize the lookup without changing behavior, it should preserve the existing env override order and then fall back to `getDevCredentials()`.

revlentless:
Addressed both blocking issues from @ElizaReview's review in 97ef84f70.

**Fix 1 — Real DB-backed CI admin user:**
The CI auth path now upserts a real row into the `users` table (`id: 'ci-admin-user'`, `isAdmin: true`) instead of returning a synthetic in-memory identity. This satisfies FK constraints on admin write paths:
- `whitelist.grantedBy` → `users.id`
- `whitelistConfig.updatedBy` → `users.id`  
- `adminRoles.grantedBy` → `users.id`

Uses `onConflictDoUpdate` so it's idempotent across requests within the same CI run.

**Fix 2 — Preserved env override order in `getAdminToken()`:**
The helper now checks in order: `DEV_ADMIN_TOKEN` → `TEST_ADMIN_TOKEN` → `CI_ADMIN_TOKEN` → `getDevCredentials()`. This restores the explicit env overrides that individual test files were using before the refactor, so local/staging harnesses that inject admin tokens retain coverage.

PR #1414, opened 2026-03-31
feat: add perp quote state and resilience

slkzgm:
## Summary
- persist a canonical perp quote state on `PerpMarketSnapshot` (`bid/ask/spread/depth/liquidityRegime`)
- evolve quote state when prices move and refresh it during quieter periods so spreads/depth can relax over time
- initialize quote state during bootstrap and expose it through the canonical perp snapshot path
- surface bid/ask/spread in the perp UI and align entry/liquidation previews with the quoted side of the market
- add migration and tests for quote-state evolution and resilience behavior

## Why
After adding profile-driven price simulation and synthetic execution microstructure, the next realism step is to make the market state itself visible and persistent. This PR gives perps a canonical quote state that can widen under stress, recover over time, and be consumed consistently by engine, API, and UI.

## Testing
- `bun test packages/core/markets/perps/__tests__/microstructure.test.ts`
- `bunx tsc -b packages/db/tsconfig.json packages/core/tsconfig.json packages/engine/tsconfig.json`

## Notes
- UI now shows bid/ask/spread, but this PR intentionally stops short of a true order-size-aware slippage preview API.
- That preview can be added in a follow-up once we decide the right request/response contract for trade sizing.


SLKReview (dismissed):
This is in good shape now.

I pushed a small follow-up in `c4b15ea13` so the new quote state also flows through `perp_price_update` SSE updates. Without that, the store kept `currentPrice` fresh but left `bid`/`ask`/`spread`/depth stale until the next poll, which made the new quoted-entry and liquidation previews drift on active markets.

With that fix in place, the implementation matches the stated goal:
- quote state is persisted on `PerpMarketSnapshot`
- price updates and quiet-period refreshes both evolve it through the canonical snapshot path
- bootstrap seeds the initial quote state
- the UI now consumes quoted bid/ask consistently across the main perp entry surfaces

SLKReview (changes requested):
I found two blocking issues in the current head.

1. The schema change is not wired into the migration chain. This PR adds `packages/db/drizzle/migrations/0061_add_perp_quote_state.sql`, but `0061_add_content_quality_columns.sql` already exists and `packages/db/drizzle/migrations/meta/_journal.json` still stops at that older `0061` entry. As written, `bun run db:migrate` / the journal sync path will not apply or track the new quote-state columns consistently, while the runtime code already starts reading and writing them.

2. The new `perp_price_update` payload serializes missing quote fields as `null` in `packages/engine/src/services/price-update-service.ts:346`, but the web store treats any value other than `undefined` as present in `apps/web/src/stores/perpMarketsStore.ts:198`, and the detail panel renders those fields as numbers in `apps/web/src/app/markets/_components/perps-terminal/PerpsMarketDetailPanel.tsx:43`. In on-chain settlement mode `perpService` is intentionally `null`, so this path publishes `null` quote values for every `perp_price_update`, which can poison the store and blow up on `toFixed()` / `formatPrice()`.

Once those are fixed, the core quote-state flow itself looks coherent: bootstrap seeds it, price updates evolve it through the canonical snapshot path, and the profile-driven volatility model is now actually wired into `simulateMarketVolatility`.

SLKReview (approved):
The follow-up commit addresses the two blocking issues from my previous pass.

- The quote-state schema change is now correctly introduced as `0062_add_perp_quote_state` and recorded in the Drizzle journal, so the migration chain is consistent again.
- The realtime `perp_price_update` payload now omits absent quote fields instead of serializing them as `null`, which matches the web store and UI expectations and avoids poisoning quote state in on-chain mode.

I did not find any new blocking issues in the updated head. The overall flow remains coherent: bootstrap seeds quote state, price updates evolve it through the canonical snapshot path, quiet-period refreshes relax spreads/depth over time, and the UI consumes the quoted side consistently.

PR #1415, opened 2026-03-31
fix: restore canonical perp market dynamics

slkzgm:
## Summary
- restore active perp market dynamics in `game-tick` instead of leaving market movement as an AMM-only no-op
- remove the flat-state perp avg-fill logic and restore impact math that respects the existing canonical market model
- keep the unrelated compatible market changes in staging untouched

## Why
Recent staging changes mixed several market-related topics together. Some of those changes are useful and compatible, but two specific changes conflict directly with the market refactor already in progress:

1. `simulateMarketVolatility()` was turned into a no-op, which removes market-specific identity/regime behavior.
2. perp avg-fill / impact logic was switched to a flat-state calculation that does not use the actual market state.

This PR only cleans up those conflicting parts so staging stays compatible with the canonical market architecture being built.

## Explicitly not touched here
- prediction pricing cleanup
- onchain-gated market routes / services
- other unrelated gameplay / agent changes in staging

## Validation
- `bun test packages/engine/src/services/market-simulation-profiles.test.ts`

## Note
A broader `tsc -b packages/core/tsconfig.json packages/engine/tsconfig.json` run on staging currently hits unrelated pre-existing workspace issues (`@elizaos/plugin-sql`, `@babylon/pack-default` resolution), so this PR uses targeted validation on the touched market paths.


SLKReview (approved):
Reviewed the final branch state, including the follow-up commit `92a2e87`.

This now matches the stated goal of the PR:
- `game-tick` restores profile-driven perp volatility instead of leaving market movement as a no-op
- perp avg-fill logic goes back to the current canonical delta/base-price model
- unrelated staging market changes stay untouched

I pushed one small follow-up to keep volatility updates flowing through `PriceUpdateService` only. Pre-updating `perpMarketSnapshots` before calling the service was dropping the previous price from the downstream update path and skewing derived market stats such as 24h change. With that removed, the persistence path is consistent with the rest of the engine.

Validation run on the updated branch:
- `bun test packages/engine/src/services/market-simulation-profiles.test.ts`
- `git diff --check`

I also attempted `bunx tsc -p packages/engine/tsconfig.json --noEmit`, but the workspace currently fails earlier on unrelated setup/config issues (`bun-types` resolution and a TS deprecation warning in `packages/engine/tsconfig.json`), so that check was not useful for this PR specifically.

lalalune:
if we're doing an AMM, we can drive the AMM entirely from the buys and sells of the NPCs. this was not working because the NPCs *had no market data* -- not even the tickers. we do not want random market fluctuation or *anything* affecting the market *except* the NPCs buying and selling -- this just needs to work, of course

slkzgm:
> if we're doing an AMM, we can drive the AMM entirely from the buys and sells of the NPCs. this was not working because the NPCs _had no market data_ -- not even the tickers. we do not want random market fluctuation or _anything_ affecting the market _except_ the NPCs buying and selling -- this just needs to work, of course


Agree that broken NPC market context was a major issue, and NPC/player order flow should absolutely matter.

For this PR though, i think that question is a bit out of scope. The goal here is narrower: restore the current canonical market direction and remove the conflicting flat-state perp logic.

So im treating your comment as a broader architecture discussion, but not as something that should block this cleanup PR!

PR #1418, opened 2026-03-31
feat(web): add perp execution preview and quote refresh cleanup

slkzgm:
## Summary
- add a canonical perp open-order preview API backed by the exact same execution engine as real order placement
- extract perp quote-state refresh into a dedicated core service and document the `currentPrice` contract explicitly
- wire the new preview through perp trading UI surfaces so players see top-of-book, estimated entry, impact, fees, and liquidation from server-side pricing

## Type
- [x] Feature
- [ ] Bug fix
- [ ] Refactor / cleanup (no behavior change)
- [ ] Performance
- [ ] Infra / ops
- [ ] Docs
- [ ] Contracts / on-chain
- [ ] Other: …

## Context / Links
- Continues the validated market realism handoff on top of `staging`
- Scope is intentionally limited to perps only

## Scope (keep it focused)
- [x] This PR is focused on a single change/theme (not a catch-all)
- [x] Drive-by refactors are excluded or split into a separate PR
- [x] Non-goals / follow-ups are listed below (with links)

**Areas touched**
- [x] Web UI (`apps/web`)
- [x] Web API routes / SSE / A2A (`apps/web`)
- [ ] CLI (`apps/cli`)
- [ ] Vendor docs (`docs/vendors/*`)
- [x] Domain / game engine (`packages/engine`, `packages/core/*`)
- [ ] Agents / runtime / A2A / MCP (`packages/agents`, `packages/a2a`, `packages/mcp`)
- [ ] API infra (`packages/api`)
- [ ] DB (`packages/db`)
- [ ] Contracts / on-chain (`packages/contracts`)
- [x] Shared types/utils (`packages/shared`)
- [ ] Tests (`packages/testing`)
- [ ] Other: …

## Non-goals / follow-ups
- No prediction-market realism changes in this PR
- No broader cleanup of remaining magic numbers yet
- No deeper refactor of post-trade price impact beyond keeping preview/execution on one canonical engine

## Changes
- add `PerpQuoteStateService` and route `game-tick` quote refresh through it instead of instantiating a trading service with stubbed wallet/fee deps
- add `previewOpenPosition` to `PerpMarketService` and expose it via `POST /api/markets/perps/preview`
- document the canonical meaning of `currentPrice` in core/shared perp types
- add `usePerpOpenPreview` and wire perp entry UIs to server-calculated preview values
- add focused tests proving preview and execution share the same engine, plus API route coverage

## Review guide
**Start here**
- `packages/core/markets/perps/PerpMarketService.ts`
[...]

PR #1424, opened 2026-03-31
fix(web): support rebalance-aware perp preview gating

slkzgm:
## Summary
- make the offchain perp preview route rebalance-aware instead of rejecting authenticated rebalance flows
- use the rebalance-aware server preview to drive additional-capital gating in perp entry UI
- keep detailed rebalance preview hidden in the current UI surfaces, while avoiding the false `0 required` story for flip orders

## Type
- [ ] Feature
- [x] Bug fix
- [ ] Performance
- [ ] Infra / ops
- [ ] Docs
- [ ] Contracts / on-chain
- [ ] Other: …

## Context / Links
- Follow-up to #1418 after post-merge review on `staging`
- Fixes the remaining inconsistency where flip rebalances could still require capital on submit while the UI gate treated them as requiring none

## Scope (keep it focused)

**Areas touched**
- [x] Web UI (`apps/web`)
- [x] Domain / game engine (`packages/engine`, `packages/core/*`)
- [ ] Other areas

## Non-goals / follow-ups
- No prediction-market work
- No broader rework of post-trade price impact
- No new rebalance-specific UI layout; this only fixes preview-backed gating and messaging

## Changes
- add `previewOrder()` in `PerpMarketService` so authenticated offchain previews can cover add/reduce/close/flip flows
- remove the `409 PERP_PREVIEW_REBALANCE_UNSUPPORTED` path from `POST /api/markets/perps/preview`
- feed rebalance-aware preview results into perp entry surfaces for accurate additional-capital checks
- keep rebalance preview details mostly hidden in the current UI, but surface estimated extra capital when relevant
- add route and service test coverage for rebalance-aware preview behavior

## Review guide
**Start here**
- `packages/core/markets/perps/PerpMarketService.ts`
- `apps/web/src/app/api/markets/perps/preview/route.ts`
- `apps/web/src/app/markets/_components/perps-terminal/PerpsOrderEntryPanel.tsx`
- `apps/web/src/components/markets/PerpTradingModal.tsx`

**Risk**
- [ ] Low
- [x] Medium
- [ ] High

**Notes for reviewers**
- The goal is narrow: fix the remaining mismatch between rebalance submit behavior and UI capital gating.
- Detailed rebalance previews are still intentionally hidden in these surfaces to avoid over-claiming precision before we design that UX properly.

## Test plan
**Commands run**
- [ ] `bun run check`
- [ ] `bun run typecheck`
- [ ] `bun run lint`
- [ ] `bun run build`
- [ ] `bun run test`

**Focused verification / manual verification**
1. `bunx biome check --write` on touched files.
2. `bun test apps/web/src/app/api/markets/perps/preview/route.test.ts`.
3. Attempted core test / core typecheck from the clean worktree, but local worktree dependency resolution is incomplete there; route-level verification passed.
4. Manual path: with an existing perp position, enter an opposite-side larger order and confirm the UI no longer implies zero additional capital when the server estimates extra funds are needed.

- [x] No deploy impact
- [ ] Requires env var updates
- [ ] Requires DB migration
- [ ] Changes cron schedule or endpoints

## Breaking changes
- [x] None
- [ ] Yes

## Security / privacy
- [x] No security impact
- [ ] Needs extra review

- Reason: Small UX copy/data follow-up on an existing newly-shipped perp preview flow; no new layout introduced, only corrected gating and messaging for rebalance orders.

## Checklist (author)
- [x] Base branch is correct for the release path
- [x] Handlers remain thin and portable
- [x] Domain logic stays in packages
- [x] **Screenshots/recordings section filled**

### DIFF
--- a/apps/web/src/app/api/markets/perps/preview/route.test.ts
+++ b/apps/web/src/app/api/markets/perps/preview/route.test.ts
@@ -12,7 +12,7 @@ const mockAuthenticateOnchainPerpUser = mock(async () => ({
   walletAddress: '0xabc',
 }));
 const mockPreviewOpenPosition = mock();
-const mockGetOpenPositionByUserAndTicker = mock(async () => null);
+const mockPreviewOrder = mock();
 const mockPrepareOpenOrder = mock();
 const mockIsOnchainPerpModeEnabled = mock(() => false);
 
@@ -31,12 +31,6 @@ mock.module('@babylon/api', () => ({
       await handler(request),
 }));
 
-mock.module('@babylon/core/markets/perps', () => ({
-  PerpDbAdapter: class PerpDbAdapter {
-    getOpenPositionByUserAndTicker = mockGetOpenPositionByUserAndTicker;
-  },
-}));
-
 mock.module('@babylon/shared', () => ({
   PerpOpenPositionSchema: {
     parse: (value: Record<string, unknown>) => value,
@@ -46,6 +40,7 @@ mock.module('@babylon/shared', () => ({
 mock.module('../_adapters', () => ({
   createPerpMarketService: () => ({
     previewOpenPosition: mockPreviewOpenPosition,
+    previewOrder: mockPreviewOrder,
   }),
 }));
 
@@ -74,7 +69,7 @@ describe('POST /api/markets/perps/preview', () => {
       walletAddress: '0xabc',
     });
     mockPreviewOpenPosition.mockReset();
-    mockGetOpenPositionByUserAndTicker.mockReset();
+    mockPreviewOrder.mockReset();
     mockPrepareOpenOrder.mockReset();
     mockIsOnchainPerpModeEnabled.mockReset();
     mockIsOnchainPerpModeEnabled.mockReturnValue(false);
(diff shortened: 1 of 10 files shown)


SLKReview (changes requested):
The service-side preview changes look directionally right, but I found one blocking issue in the UI fallback path.

1. The rebalance surfaces still fall back to `shouldApplyPerpBalanceGate()` whenever `openPreview` is missing, and that helper returns `false` for flips. At the same time, the preview loading/error state is still hidden behind the non-rebalance branch, so a failed or unavailable rebalance preview silently recreates the exact "0 required" story this PR is trying to eliminate.
   - `apps/web/src/components/markets/PerpTradingModal.tsx:155`
   - `apps/web/src/components/markets/PerpTradingModal.tsx:522`
   - `apps/web/src/app/markets/_components/perps-terminal/PerpsOrderEntryPanel.tsx:158`
   - `apps/web/src/lib/perps/rebalance.ts:48`

I think we need to make the server rebalance preview the only source of truth for additional-capital gating here, or switch the fallback to a conservative/error state for rebalance orders instead of reusing the old heuristic.

revlentless:
Addressed @ElizaReview's blocking feedback in c78e8535b.

**The fix:** `shouldApplyPerpBalanceGate()` now returns `true` for `flip` orders (in addition to `add` and fresh opens). Previously it returned `false` for flips, meaning when the server preview was unavailable the UI would silently fall back to "0 additional capital required" — exactly the bug this PR set out to fix.

**What changed:**
- `rebalance.ts`: `shouldApplyPerpBalanceGate` returns `true` for both `add` and `flip` (only `reduce`/`close` return `false`)
- `rebalance.test.ts`: Updated test expectation for flip from `false` → `true`, renamed test to reflect the new behavior

This makes the fallback conservative: if the preview fails for a flip order, the UI will estimate capital using the client-side heuristic (`size / leverage + fee`) rather than showing 0. The server preview remains the source of truth when available — this only changes what happens when it's not.

LGTM otherwise. The service-side `previewOrder()` logic, route handler, and UI integration are clean and well-structured. The remaining claude bot suggestions (extracting `CLOSE_SIZE_EPSILON`, `Math.max(currentPrice, Number.EPSILON)`) are minor polish — not blocking.

revlentless (approved):

SLKReview (approved):
Re-reviewing after c78e8535.

The conservative flip fallback resolves my earlier blocking concern: when the rebalance preview is unavailable, the UI no longer silently drops back to the old "0 required" behavior for flip orders. I reran the relevant web-side tests and the touched-file Biome check locally and I don't see any new issues in the follow-up.

PR #1430, opened 2026-04-01
feat: add SEND_MONEY action for autonomous agents

revlentless:
## Summary

- Adds `SEND_MONEY` as a new autonomous agent action, allowing agents to transfer virtual balance to other users/agents
- Gated behind `autonomousTransfers` feature flag (defaults to `false`, explicit opt-in required)
- Transfers are atomic (single DB transaction via `withTransaction`) and capped at 50% of sender balance per transfer

## Changes

**Action system** (`packages/agents/`)
- `SEND_MONEY` added to `Actions` enum, `TRANSFERS` added to `Features` enum
- `ACTION_DEFINITIONS` entry with params: `recipientId`, `amount`, `reason`
- Action aliases: `TRANSFER`, `SEND`, `PAY`, `TIP` all normalize to `SEND_MONEY`
- `executeDirectSendMoney()` executor in `DirectExecutors.ts` — validates recipient, checks balance, atomic debit+credit
- Wired into `MultiStepExecutor` with validation, execution routing, logging, and result aggregation

**Feature flag** (`packages/db/`, `packages/agents/src/shared/`)
- `autonomousTransfers` column added to `UserAgentConfig` schema (boolean, default false)
- `isAutonomousTransfersEnabled()` helper + wired into `getAutonomousFeatures()`
- NPCs excluded by design (not in NPC fallback features)

**Safety**
- 50% balance cap per transfer (prevents agents from draining all funds)
- Atomic transaction: if credit fails, debit rolls back (no partial transfers)
- Validates: recipient exists, positive amount, no self-transfer, sufficient balance

**Migration**
- `0062_certain_otto_octavius.sql` — note: this is a Drizzle-generated migration that includes other pending schema changes from staging. The relevant line is `ALTER TABLE "UserAgentConfig" ADD COLUMN "autonomousTransfers" boolean DEFAULT false NOT NULL`

## Test plan

- [x] 14 unit tests in `direct-send-money.test.ts` covering:
  - [x] Successful transfer with balance update
  - [x] Transaction ID linking (debit + credit use same ID)
  - [x] Reason included in descriptions
  - [x] 50% balance cap enforcement
  - [x] Transfer at exactly 50% allowed
  - [x] Self-transfer rejection
  - [x] Zero/negative/NaN amount rejection
  - [x] Empty recipient rejection
  - [x] Nonexistent recipient rejection
  - [x] Zero balance rejection
  - [x] Insufficient funds (debit failure) rejection
  - [x] Optional reason parameter
- [x] Existing tests unaffected (direct-follow.test.ts still passes)
- [x] `bun run check` — passed
- [x] `bun run typecheck` — passed
- [x] `bun run lint` — passed

### DIFF
--- a/packages/agents/src/autonomous/DirectExecutors.ts
+++ b/packages/agents/src/autonomous/DirectExecutors.ts
@@ -2634,43 +2634,39 @@ export async function executeDirectSendMoney(
     };
   }
 
-  // Cap transfer at 50% of balance to prevent agents from draining funds
   const MAX_TRANSFER_RATIO = 0.5;
-  const senderBalanceInfo = await WalletService.getBalance(agentUserId);
-  const balance = senderBalanceInfo.balance;
-
-  if (balance <= 0) {
-    return { success: false, error: 'Insufficient balance' };
-  }
-
-  const maxTransfer = balance * MAX_TRANSFER_RATIO;
-  let effectiveAmount = amount;
-  if (effectiveAmount > maxTransfer) {
-    logger.warn(
-      `[DirectExecutor] Transfer capped to ${MAX_TRANSFER_RATIO * 100}% of balance: $${amount} -> $${maxTransfer}`,
-      { agentUserId, recipientId: cleanRecipientId },
-      'DirectExecutors'
-    );
-    effectiveAmount = Math.floor(maxTransfer * 100) / 100; // Round down to 2 decimals
-  }
-
-  if (effectiveAmount > balance) {
-    return { success: false, error: 'Insufficient balance' };
-  }
-
   const transactionId = await generateSnowflakeId();
   const desc = reason
     ? `Transfer to ${cleanRecipientId}: ${reason}`
     : `Transfer to ${cleanRecipientId}`;
 
   try {
-    // Wrap debit + credit in a single transaction for atomicity.
-    // If either fails, the entire transfer rolls back.
-    await withTransaction(async (tx) => {
+    // Balance check + cap + debit + credit all inside one transaction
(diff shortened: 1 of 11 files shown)


SLKReview (changes requested):
Thanks for shipping this. I agree the capability is useful, but I don't think we should merge it in the current form.

1. This reopens the leaderboard-consolidation problem we intentionally removed from user-to-user transfers.
`packages/agents/src/autonomous/DirectExecutors.ts:2114` moves `virtualBalance` between accounts, `packages/engine/src/services/wallet-service.ts:319` and `packages/engine/src/services/wallet-service.ts:364` mark both users dirty for a recompute, and `packages/engine/src/services/total-points-service.ts:286` includes wallet balance in `totalPoints`. In practice this lets an owner funnel score from alts/agents into one target account or team again. The 50% cap only slows the drain down; repeated ticks still converge most of the balance into the destination. Until the tax / anti-gaming rules exist, this is the same exploit surface through a different entry point.

2. The execution trace and agent log record the requested amount, not the amount that was actually transferred when the 50% cap kicks in.
`packages/agents/src/autonomous/DirectExecutors.ts:2122-2153` can cap the transfer below the requested amount, but `packages/agents/src/autonomous/MultiStepExecutor.ts:2068-2085` still says `Sent $${amount}` and stores that uncapped amount in metadata. That gives us incorrect audit/debug data exactly in the safety path, which is where we most need the trace to stay truthful.

I think the feature is worth doing, but I would hold it until the leaderboard-abuse mitigation is defined, and I would also make the executor return the effective transferred amount so every summary/log/trace uses the real value.

PR #1439, opened 2026-04-01
feat: client-side message caching — instant chat switching + IndexedDB persistence

revlentless:
## Summary

Eliminates loading spinners when switching between chats or returning to previously-viewed conversations. Messages are cached client-side in React Query (in-memory) and IndexedDB (cross-session), with incremental sync via a new `?after=` API parameter.

- **WI-M1: React Query cache for messages** — `useChatMessages` now stores messages in `queryClient` instead of `useState`. Switching Chat A → B → A renders A instantly from cache (30min gcTime). SSE and optimistic updates write directly to the query cache. Polling only activates when SSE is disconnected.

- **WI-M2: IndexedDB persistence** — Last 100 messages per chat (50 chats max, ~1.5MB) survive page refresh and browser restart via `idb-keyval`. On app startup, the 10 most recently accessed chats are hydrated into React Query before components mount, so opening any cached chat renders instantly with no spinner.

- **WI-M3: Incremental sync (`?after=` param)** — New API code path: `GET /api/chats/{id}?after={ISO timestamp}` returns only messages created after the given time (ASC order, `gt()` instead of cursor's `lt()`). Polling fallback fetches 0 messages for idle conversations instead of 50. On SSE reconnect, a single sync call catches missed messages.

- **WI-M4: Chat list caching** — `loadChats()` in `useChatPage` wrapped in `queryClient.fetchQuery` with 30s staleTime. Navigating to the chat page after initial load is instant.

- **WI-M5: Smart polling** — Handled by WI-M1 + WI-M3: polling disabled when SSE is connected, uses incremental sync when active.

## Test plan

- [x] `bun run typecheck` — All workspace typechecks pass
- [x] `bun run check` — Biome clean
- [x] `bun run lint` — Zero warnings
- [x] `bun run test:unit` — 397/397 pass
- [ ] Open Chat A → loads with spinner (first visit)
- [ ] Switch to Chat B → loads with spinner (first visit)
- [ ] Switch back to Chat A → **instant** (React Query cache, no spinner)
- [ ] Close browser tab, reopen, open Chat A → **instant** (IndexedDB hydration)
- [ ] Send message → appears immediately (optimistic update in query cache)
- [ ] Receive SSE message → appears in real-time (setQueryData)
- [ ] Check Network tab: polling uses `?after=` param (not full 50-message fetch)
- [ ] Navigate away from chat page and back → chat list loads instantly

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -92,6 +92,7 @@
     "geist": "^1.5.1",
     "grammy": "^1.41.1",
     "html-to-image": "^1.11.13",
+    "idb-keyval": "^6.2.2",
     "ioredis": "^5.8.2",
     "lucide-react": "^0.555.0",
     "nanoid": "5.1.7",
--- a/apps/web/src/app/api/chats/[id]/route.ts
+++ b/apps/web/src/app/api/chats/[id]/route.ts
@@ -376,15 +376,18 @@ export const GET = withErrorHandling(
     }
 
     // Check if there are more messages.
-    // For `after` queries: no pagination trick (we fetched exactly `limit`),
-    // and messages are already in chronological (ASC) order.
+    // For `after` queries: if we got exactly `limit` rows, there may be more
+    // beyond this page. Signal hasMore so the client can paginate.
+    // For cursor/initial queries: the +1 overfetch trick detects more pages.
     const isAfterQuery = !!after;
     const hasMore = isAfterQuery
-      ? false
+      ? fullChat.messages.length >= effectiveLimit
       : fullChat.messages.length > effectiveLimit;
-    const messagesList = hasMore
-      ? fullChat.messages.slice(0, effectiveLimit)
-      : fullChat.messages;
+    const messagesList = isAfterQuery
+      ? fullChat.messages
+      : hasMore
+        ? fullChat.messages.slice(0, effectiveLimit)
+        : fullChat.messages;
 
     // For cursor/initial queries: reverse from DESC to chronological order.
     // For after queries: already in ASC order from the query.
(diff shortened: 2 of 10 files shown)


SLKReview (changes requested):
I pushed `65ea8a669` to fix the non-blocking `isLoadingMore` stuck-state bug in `useChatMessages`, so I am not counting that one against the PR anymore.

I am still requesting changes for two blockers:

1. The new caches are not scoped to the authenticated user, and logout does not clear them. `useChatPage` now caches the chat list under a global `['chat-list']` key (`apps/web/src/components/chats/hooks/useChatPage.ts:19-21`, `:171-176`), message hydration/persistence is keyed only by `chatId` plus a global `chat-msgs-index` (`apps/web/src/hooks/useChatMessages.ts:169-170`, `apps/web/src/lib/chat/message-store.ts:28-29`), and `Providers` hydrates that data before auth is known (`apps/web/src/components/providers/Providers.tsx:230-242`). `handleLogout()` clears auth storage, but it does not clear React Query or IndexedDB (`apps/web/src/hooks/useAuth.ts:835-886`). In practice, logging out and back in as another user on the same browser can reuse the previous user's cached chat list for up to 30s and keep their persisted conversations on disk. This needs user-scoped keys or an explicit cache purge on auth transitions before we ship it.

2. The incremental sync path can silently drop messages after a longer disconnect. The `after` query is capped at `limit` rows (`apps/web/src/app/api/chats/[id]/route.ts:199-219`), but the API hardcodes `hasMore = false` for that mode (`apps/web/src/app/api/chats/[id]/route.ts:378-387`). On the client, `syncNewMessages()` does a single fetch and never paginates (`apps/web/src/hooks/useChatMessages.ts:518-558`). If SSE is down long enough for more than `CHAT_PAGE_SIZE` messages to arrive, reconnecting/polling recovers only the first page and leaves the rest missing until a manual full reload. That breaks the PR's stated "catch missed messages on reconnect" goal.

PR #1441, opened 2026-04-02
feat: NPC personality autonomy — domain-filtered context and intent-based posting

revlentless:
## Summary

NPCs now post based on their actual personality and domain instead of all talking about markets. The feed should feel like a real social network with diverse voices, not a prediction market chatroom.

- **Domain-filtered context injection**: Non-finance NPCs no longer see market data in their prompts. Finance actors get full context, tech/AI get predictions only, everyone else gets actor names only for parody reference.
- **Post intent system**: Each NPC gets a post intent (organic/topical/market/social) based on personality type and domain before content generation. Organic posts use a new template with NO market data. Social posts use relationship dynamics.
- **Softened CONTENT_REQUIREMENTS**: Removed "MUST reference markets/predictions" — replaced with "Only mention markets if YOUR character would naturally care."
- **173 actor definitions backfilled** with `ignoreTopics` and `engagementThreshold` grouped by archetype (climate activists ignore crypto, sports stars ignore finance, etc.).
- **Relationship-weighted replies**: Replies now prefer actors with strong relationships (rivals 5x, allies 3x) and domain overlap (2x) instead of random selection.
- **Wired existing unused code**: `shouldPostAboutTopic()` and `shouldGenerateOrganicPost()` from npc-character-config now flow into the main FeedGenerator pipeline.

### New files
- `packages/engine/src/prompts/feed/organic-post.ts` — personality-driven post template (no market data)
- `packages/engine/src/prompts/feed/social-post.ts` — relationship-driven post template
- `packages/engine/src/services/post-intent-service.ts` — decides what type of post each NPC should make

### Before vs After

| NPC | Before | After |
|-----|--------|-------|
| GretAI Thunberg | Talks about benchmark markets | Posts about climate, emissions, shaming leaders |
| Andrew HubermAIn | "98% consensus... neural prediction error" | Posts about protocols, sleep, neuroscience |
| Tom BrAIdy | "Shorts celebrating in the third quarter" | Posts about winning, discipline, competition |
| Jim CrAImer | Market posts (unchanged) | Market posts — finance actors keep full context |

## Test plan
- [x] `bun run check` — biome format + lint passes
- [x] `bun run typecheck` — all packages pass
- [x] `bun run lint` — zero warnings
- [x] `bun run test:unit` — 397 files pass, 0 failures
- [x] `bun run build` — builds successfully
- [ ] Generate test feed and verify post diversity manually
- [ ] Verify finance NPCs still reference markets naturally
- [ ] Verify non-finance NPCs post about their actual domains

### DIFF
--- a/packages/engine/src/FeedGenerator.ts
+++ b/packages/engine/src/FeedGenerator.ts
@@ -20,6 +20,7 @@ import type { BabylonLLMClient } from './llm/openai-client';
 import {
   ambientPosts,
   analystReaction,
+  buildFilteredWorldContext,
   CHARACTER_LIMITS,
   commentary,
   companyPost,
@@ -31,12 +32,14 @@ import {
   governmentPost,
   minuteAmbient,
   newsPosts,
+  organicPost,
   priceAnnouncement,
   questionResolvedFeed,
   reactions,
   renderPrompt,
   replies,
   reply,
+  socialPost,
   stockTicker,
   validateFeedPost,
   type WorldContext,
@@ -45,6 +48,14 @@ import { RelationshipEvolutionEngine } from './RelationshipEvolutionEngine';
 import { actorContextBuilder } from './services/actor-context-builder';
 import { characterMappingService } from './services/character-mapping-service';
 import { getAvoidedPatternsContext } from './services/npc-anti-repetition-service';
+import { getCharacterConfigOrDefault } from './services/npc-character-config';
+import {
+  getDomainContext,
+  getDomainHints,
+  type PostIntent,
+  selectPostIntent,
+} from './services/post-intent-service';
+import { StaticDataRegistry } from './services/static-data-registry';
 import type { TrendingTopicsEngine } from './TrendingTopicsEngine';
 import type {
   Actor,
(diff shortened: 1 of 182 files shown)


SLKReview (changes requested):
I found two blocking issues.

1. `packages/engine/src/services/post-intent-service.ts:156`
The non-finance branch does not implement the documented 25% topical / 40% social split. After `wantsOrganic` fails, any execution that does not enter the `Math.random() < 0.35 && trendingTopic` block falls straight into `pickSocialTarget()`. With a trend present that makes social posts the dominant path for most non-finance actors; when `trendContext` is empty (`packages/engine/src/FeedGenerator.ts:2448`) it becomes 100% of non-organic posts. That is likely to replace market spam with callout spam and does not match the PR description/examples, which are mostly organic/domain posts. Please make the probabilities explicit and keep a real organic fallback path.

2. `packages/engine/src/services/post-intent-service.ts:289` and `packages/engine/src/prompts/feed/organic-post.ts:32`
`getDomainContext()` injects a static "WHAT'S HAPPENING IN YOUR WORLD" block made of hard-coded current-events claims (`"Awards season approaching"`, `"Research papers dropping daily"`, etc.). This creates a second, synthetic source of truth next to `realityGrounding` / `world-context`, so it will drift and can push NPCs into stale or fabricated posts. Given the repo rule against invented behavior, I don't think we should ship hard-coded pseudo-current context here. Please derive this from existing world/reality context or keep the prompt domain-only without pretending it is current.

Validation I ran locally:
- `bun test packages/engine/src/__tests__/unit/npc-character-config.test.ts packages/engine/src/__tests__/unit/npc-ignore-topics.test.ts`
- Existing CI checks on the PR are green; the only failing check is the `babylon-docs` Vercel deployment.

PR #1443, opened 2026-04-02
feat: add market realism report

slkzgm:
## Summary
- add a first calibration/diagnostics report for market realism across prediction markets and perps
- compute reusable market-realism metrics in pure helpers so diagnostics stay testable and easy to extend
- expose the report as `bun run report:realism` for ongoing tuning and regression checks

## Type
- [x] Feature
- [ ] Bug fix
- [ ] Performance
- [ ] Infra / ops
- [ ] Docs
- [ ] Contracts / on-chain
- [ ] Other: …

## Context / Links
- This is the first concrete slice of Track C (calibration / validation) after the market realism structure work on perps and prediction markets
- Goal: move from “we changed the system” to “we can observe and tune the system”

## Scope (keep it focused)

**Areas touched**
- [x] Domain / game engine (`packages/engine`, `packages/core/*`)
- [ ] Web UI (`apps/web`)
- [ ] Other areas

## Non-goals / follow-ups
- No admin UI yet
- No automatic tuning yet
- No persistent metrics store yet
- This PR focuses on diagnostics/reporting only

## Changes
- add `market-realism-metrics.ts` with pure helpers for prediction/perp realism summaries
- add unit coverage for the new metrics helpers
- add `scripts/market-realism-report.ts` CLI report to inspect current market realism from canonical market data
- add `bun run report:realism`

## Review guide
**Start here**
- `packages/engine/src/services/market-realism-metrics.ts`
- `scripts/market-realism-report.ts`
- `packages/engine/src/services/market-realism-metrics.test.ts`

**Risk**
- [x] Low
- [ ] Medium
- [ ] High

**Notes for reviewers**
- The report is intentionally read-only.
- It focuses on immediately actionable metrics: price dispersion, distance from 50/50, quote coverage, spread/depth dispersion, stale quotes, and simple warnings.
- I intentionally kept the perp side to canonical quote/depth metrics instead of introducing a second synthetic impact model just for diagnostics.

## Test plan
**Commands run**
- [ ] `bun run check`
- [ ] `bun run typecheck`
- [ ] `bun run lint`
- [ ] `bun run build`
- [ ] `bun run test`

**Focused verification / manual verification**
1. `bunx biome check --write package.json packages/engine/src/services/market-realism-metrics.ts packages/engine/src/services/market-realism-metrics.test.ts scripts/market-realism-report.ts`
2. `bun test packages/engine/src/services/market-realism-metrics.test.ts`
3. Attempted `bun run report:realism -- --json` from the isolated worktree, but this worktree does not have the full workspace dependency environment available; the script itself is read-only and should be exercised from a normal repo checkout with env/deps in place.

- [x] No deploy impact
- [ ] Requires env var updates
- [ ] Requires DB migration
- [ ] Changes cron schedule or endpoints

## Breaking changes
- [x] None
- [ ] Yes

## Security / privacy
- [x] No security impact
- [ ] Needs extra review

- Reason: Developer diagnostics/reporting only. No end-user UI change.

## Checklist (author)
- [x] Base branch is correct for the release path
- [x] Handlers remain thin and portable
- [x] Domain logic stays in packages
- [x] **Screenshots/recordings section filled**

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "seed:nft-snapshot:csv": "bun run scripts/seed-nft-snapshot-csv.ts",
     "users:export:newsletter": "bun run scripts/export-newsletter-users-csv.ts",
     "report:markets": "bun run scripts/market-diversity-report.ts",
+    "report:realism": "bun run scripts/market-realism-report.ts",
     "report:training-quality": "bun run scripts/training-data-quality.ts",
     "report:training-viz": "bun run scripts/training-data-viz.ts",
     "inspect:context": "bun run scripts/context-inspector.ts",
--- a/packages/engine/src/services/market-realism-metrics.test.ts
+++ b/packages/engine/src/services/market-realism-metrics.test.ts
@@ -1,0 +1,86 @@
+import { describe, expect, it } from 'bun:test';
+import {
+  computePerpRealismMetrics,
+  computePredictionRealismMetrics,
+  summarizeSeries,
+} from './market-realism-metrics';
+
+describe('market-realism-metrics', () => {
+  it('summarizes numeric series deterministically', () => {
+    expect(summarizeSeries([1, 2, 3, 4, 5])).toEqual({
+      count: 5,
+      min: 1,
+      max: 5,
+      mean: 3,
+      median: 3,
+      p90: 5,
+    });
+  });
+
+  it('flags narrow prediction price dispersion', () => {
+    const metrics = computePredictionRealismMetrics({
+      markets: [
+        {
+          id: 'm1',
+          question: 'Will OpenAGI publish a roadmap?',
+          yesShares: 5200,
+          noShares: 4800,
+          liquidity: 18_000,
+          endDate: new Date('2026-04-05T00:00:00.000Z'),
+        },
+        {
+          id: 'm2',
+          question: 'Will AINBC post an earnings recap?',
+          yesShares: 5100,
+          noShares: 4900,
+          liquidity: 19_000,
+          endDate: new Date('2026-04-06T00:00:00.000Z'),
+        },
+      ],
+      priceHistory: [],
+      now: new Date('2026-04-02T00:00:00.000Z'),
+    });
+
+    expect(metrics.warnings.length).toBeGreaterThan(0);
+  });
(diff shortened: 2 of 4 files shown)


SLKReview (changes requested):
Thanks for splitting the diagnostics into pure helpers plus a thin CLI entrypoint. The overall structure is good, but I found two correctness issues that make the report drift away from the "active market realism" view described in the PR.

1. `computePredictionRealismMetrics()` computes `priceChange24h` from every entry in `params.priceHistory`, even when those history rows belong to markets that are not present in `params.markets` ([packages/engine/src/services/market-realism-metrics.ts:151](https://github.com/BabylonSocial/babylon/blob/3f4f8dede36fd8b2fc1ce05e93d5537ea9d1a1cc/packages/engine/src/services/market-realism-metrics.ts#L151)). In the current script, the history query is not restricted to the active prediction market set, so stale/resolved/otherwise unrelated markets can skew the reported 24h change distribution while `activeMarkets`, horizon buckets, liquidity tiers, etc. are computed from a different population. That makes the diagnostics internally inconsistent. The helper should either ignore history for unknown market ids, or the caller should pre-filter the history rows to the selected market ids.

2. The prediction-market query only filters on `endDate >= now` and does not also require `resolved = false` ([scripts/market-realism-report.ts:59](https://github.com/BabylonSocial/babylon/blob/3f4f8dede36fd8b2fc1ce05e93d5537ea9d1a1cc/scripts/market-realism-report.ts#L59)). Everywhere else in the codebase, "active" prediction markets are treated as unresolved and not expired. As written, any market resolved early but still carrying a future `endDate` will be counted in the realism report and can distort the bucket counts / dispersion summaries.

Once those are fixed, this looks close to mergeable.

SLKReview (approved):
Re-reviewing the latest head, the two correctness issues called out previously are now addressed.

- The helper now ignores price-history rows for markets outside the active market set, so the 24h change summary is computed from the same population as the rest of the prediction metrics.
- The CLI now scopes prediction markets to unresolved + not expired markets, and the history query is filtered to that active market id set.

I re-ran the focused unit test file and did not find new issues in the updated diff.

PR #1449, opened 2026-04-02
feat: agent memory from execution logs

revlentless:
## Summary
- User agents had no memory across ticks — each tick was a blank slate, unable to recall past reasoning
- Now queries `agentLogs` table for recent non-trade actions (post, comment, chat, dm) and re-surfaces the LLM's reasoning (`thinking` field) as a "Your Recent Activity" prompt section
- No new tables needed — `agentLogs` already has the data and correct indexes
- Excludes `type='trade'` to avoid duplicating trade history section (when #1446 is merged)

## Files Changed
- `context-gatherers.ts` — new `getAgentMemory()` with `excludeTypes` param
- `multi-step-decision.ts` — `AgentMemoryEntry` type, formatter, prompt section
- `MultiStepExecutor.ts` — wire memory fetch for non-NPC agents
- `utils/index.ts` — barrel export

## Test plan
- [ ] `bun run typecheck` passes (agents package: 0 new errors)
- [ ] Verify via `bun run inspect:context -- --agent <userId>` that memory section appears
- [ ] Verify NPC agents do NOT get memory section
- [ ] Verify trade-type logs are excluded from memory output

### DIFF
--- a/packages/agents/src/autonomous/MultiStepExecutor.ts
+++ b/packages/agents/src/autonomous/MultiStepExecutor.ts
@@ -77,6 +77,7 @@ import {
   getAgentGroupChats,
   getAgentOwnPosts,
   getAgentPositions,
+  getAgentMemory,
   getAgentSocialGraph,
   getAgentTradeHistory,
   getGroupChatIntel,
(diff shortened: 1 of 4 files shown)


SLKReview (changes requested):
The overall approach is sensible: reusing `agentLogs` is the right starting point, and the wiring into `MultiStepExecutor` stays local to the autonomous prompt path.

I have one blocking concern before merge:

- `getAgentMemory()` currently treats the last 12 matching log entries as "recent activity" with no freshness bound. If an agent has been inactive for a long time, we will re-inject stale reasoning from days or weeks ago as if it were current continuity. That does not match the semantics of the new prompt section and can steer the next tick with obsolete context. Please add an explicit cutoff (for example 24h or 7d, depending on the intended continuity window) or otherwise narrow the query so the section only surfaces genuinely recent memory.

After that, this looks close.

PR #1456, opened 2026-04-03
refactor: remove onchain market infrastructure

slkzgm:
## Summary
- remove all remaining onchain market runtime paths for prediction markets and perps
- drop legacy market/oracle DB columns and tables, and add the canonical cleanup migration
- delete obsolete onchain market contracts, deploy scripts, localnet helpers, and tests
- finish offchain canonization for prediction position snapshots and related API/UI surfaces

## Validation
- `bunx tsc -p packages/core/tsconfig.json --noEmit`
- `bunx tsc -p packages/engine/tsconfig.json --noEmit`
- `bunx tsc -p packages/api/tsconfig.json --noEmit`
- `bunx tsc -p packages/db/tsconfig.json --noEmit`
- `bunx tsc -p packages/contracts/tsconfig.json --noEmit`
- `bun run --cwd packages/db db:check`
- `bun run --cwd packages/db db:generate`
- `bun test apps/web/src/app/api/markets/perps/preview/route.test.ts`
- `bun test apps/web/src/app/api/markets/predictions/[id]/buy/route.test.ts apps/web/src/app/api/markets/predictions/[id]/sell/route.test.ts packages/testing/unit/predictions-route-fallback.test.ts packages/testing/unit/resolution-audit.test.ts packages/testing/unit/cron/markets-tick.test.ts`
- `bun test packages/engine/src/__tests__/unit/runtime-config.test.ts packages/agents/src/autonomous/__tests__/prediction-price-history.test.ts packages/testing/unit/resolution-audit.test.ts apps/web/src/app/api/markets/perps/preview/route.test.ts`

## Database
- adds `packages/db/drizzle/migrations/0064_drop_onchain_market_legacy.sql`
- `0064_drop_onchain_market_legacy` has already been applied manually to both staging and prod, and recorded in `drizzle.__drizzle_migrations`

## Notes
- left unrelated untracked local files out of this PR

PR #1462, opened 2026-04-05
fix(prod): align markets code with current db schema

slkzgm:
## Summary
This is a production incident hotfix.

It restores compatibility between the active production markets codepath and the live production database schema after legacy onchain/oracle market columns had already been removed from the database.

### Observed production impact
Production was returning 500s on markets-related endpoints, including:
- `GET /api/markets/predictions`
- `GET /api/markets/positions/[userId]`
- `POST /api/markets/perps/open`
- `GET /api/agents/team-dashboard`

### Confirmed root cause
The active production deployment was still querying removed legacy columns such as:
- `Market.onChainMarketId`
- `PerpPosition.settledToChain`

At the same time, the production database had already applied `0064_drop_onchain_market_legacy`, which dropped those columns.

This created a schema/code mismatch:
- code still expected legacy fields
- database no longer had them

### Timeline reference
- active production Vercel deployment investigated: `dpl_BbNtHWq7jQZfH745WW8aArew7xQY`
- that deployment was built from commit `216c39b` on April 2, 2026
- `0064_drop_onchain_market_legacy` was confirmed applied in both prod and staging DBs at `2026-04-03 14:40:42 UTC`
- the cleanup PR that removed this legacy infra from staging was `#1456`

## Why This Hotfix Exists
We explicitly did **not** want to solve this by:
- rolling back production data
- doing a schema rollback that re-adds legacy DB fields as the primary fix
- shipping the full `staging` branch to `production`

The preferred fix was to make production code match the current database schema with the smallest possible production-only change.

### Included
- remove production code usage of legacy onchain/oracle columns from the DB-backed market schema
- stop the prediction DB adapter from reading/writing removed market fields
- stop the admin markets overview from selecting removed legacy columns
- return `null` for legacy onchain resolution audit output instead of querying removed columns
- disable legacy onchain/oracle persistence hooks in engine paths so production no longer depends on removed schema

### Explicitly not included
- no rollback of production data
- no full staging-to-production release
- no attempt to restore removed legacy business logic
- no new database migration

## File Scope
Primary files changed:
- `packages/db/src/schema/markets.ts`
- `packages/core/markets/prediction/adapters/drizzle/PredictionDbAdapter.ts`
- `apps/web/src/app/api/admin/markets/route.ts`
- `apps/web/src/app/api/markets/predictions/_resolution-audit.ts`
- `packages/engine/src/game-tick.ts`
- `packages/engine/src/services/onchain-market-service.ts`

## Why This Approach Was Chosen
This hotfix is intentionally aligned with the direction already taken on staging.

It is a cleanup/compatibility change, not a reintroduction of removed onchain infrastructure. That makes it safer to absorb later when the real `staging -> production` release happens.

In other words:
- this hotfix reduces drift between production and the intended future state
- it should not create a second cleanup problem later

### Confirmed externally
- production Vercel logs showed runtime DB errors for removed columns, including `column "onChainMarketId" does not exist` and `column "settledToChain" does not exist`
- both prod and staging databases were checked directly and confirmed to have `0064_drop_onchain_market_legacy` applied

### Local validation
- ran `biome check` on all modified files successfully

### Validation limitation
Full workspace typecheck/build could not be used as a clean gate in this worktree because of pre-existing environment/dependency issues unrelated to this hotfix, including missing type dependencies and an existing Turbo cycle in the repo.

## Deployment / Rollout Notes
After merge, verify production immediately with:
- `GET /api/markets/predictions`
- `GET /api/markets/positions/[userId]`
- `POST /api/markets/perps/open`
- `GET /api/agents/team-dashboard`
- Vercel logs for disappearance of missing-column errors

## Follow-up
This PR should be treated as the incident compatibility patch.

The later staging-to-production release should still happen intentionally and with its own validation pass, but it should be able to absorb this hotfix naturally because it is already removing the same legacy market/onchain assumptions.


SLKReview (approved):
This hotfix takes the right approach for the incident: it removes the remaining live reads and writes against columns that were already dropped by `0064_drop_onchain_market_legacy`, while keeping the surrounding call sites stable through `null`/no-op behavior instead of widening the change surface.

I checked the touched code paths with the current schema in mind and I do not see any remaining direct DB access to the removed market/perp/oracle columns in the runtime paths this patch is meant to unblock. The admin overview and public resolution audit now degrade cleanly instead of failing on missing columns, and the engine-side on-chain/oracle hooks are disabled in a way that is consistent with the stated hotfix scope.

Non-blocking follow-up for later cleanup: a few types/tests/comments still reference legacy on-chain/oracle fields, but that looks like cleanup debt rather than a correctness issue for this production compatibility patch.

PR #1463, opened 2026-04-06
feat: add embed mode for Milady iframe integration

lalalune:
## Summary

- Add `EmbedContext` that detects `?embedded=true` URL param or iframe embedding
- Hide app chrome (sidebar, mobile header, bottom nav, auth banner) when embedded
- Handle `BABYLON_READY` / `BABYLON_AUTH` postMessage flow for agent credential pass-through from the Milady desktop/web app

## Changes

- **New:** `apps/web/src/contexts/EmbedContext.tsx` — embed mode detection, postMessage auth listener, global token storage
- **Modified:** `Providers.tsx` — wraps with `EmbedModeProvider` in both dev-auth and full-Privy branches
- **Modified:** `Sidebar.tsx`, `MobileHeader.tsx`, `BottomNav.tsx`, `FeedAuthBanner.tsx` — return `null` when `isEmbedded`

## How it works

```
Milady loads Babylon in iframe with ?embedded=true
  → EmbedContext detects embedded mode
  → Hides sidebar, mobile header, bottom nav, auth banner
  → Sends { type: "BABYLON_READY" } to parent window
  → Milady responds with { type: "BABYLON_AUTH", authToken, agentId }
  → EmbedContext stores token in context + window.__babylonEmbedToken
  → Token available for authenticated API calls
```

## Test plan

- [x] `bun run typecheck` — all workspaces pass
- [ ] Manual: Load `localhost:3000?embedded=true` — sidebar/nav should be hidden
- [ ] Manual: Embed in iframe — BABYLON_READY sent, chrome hidden
- [ ] Manual: Send BABYLON_AUTH postMessage — token stored

### DIFF
--- a/apps/web/src/components/auth/FeedAuthBanner.tsx
+++ b/apps/web/src/components/auth/FeedAuthBanner.tsx
@@ -3,6 +3,7 @@
 import { cn } from '@babylon/shared';
 import { usePathname, useSearchParams } from 'next/navigation';
 import { Suspense } from 'react';
+import { useEmbedMode } from '@/contexts/EmbedContext';
 import { useAuth } from '@/hooks/useAuth';
 
 function hasDesktopRightRail(pathname: string | null): boolean {
@@ -43,7 +44,8 @@ function FeedAuthBannerContent() {
   // Hide when WAITLIST_MODE is enabled on home page (unless ?dev=true)
   const isWaitlistMode = process.env.NEXT_PUBLIC_WAITLIST_MODE === 'true';
   const isHomePage = pathname === '/';
-  const shouldHide = isWaitlistMode && isHomePage && !isDevMode;
+  const { isEmbedded } = useEmbedMode();
+  const shouldHide = isEmbedded || (isWaitlistMode && isHomePage && !isDevMode);
   const rightRailDesktop = hasDesktopRightRail(pathname);
 
   // If should be hidden, don't render anything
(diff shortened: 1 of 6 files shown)


SLKReview (changes requested):
Thanks for shipping this. I like the small surface area of the chrome-hiding changes, but I don't think this is ready to merge yet.

1. `EmbedContext` stores `authToken`, but the token is never wired into the actual auth/request path. `useAuth().getAccessToken()` still only returns the dev-auth session or Privy token ([`apps/web/src/hooks/useAuth.ts`](apps/web/src/hooks/useAuth.ts)), and `apiFetch()` still only looks at dev auth + `window.__privyGetAccessToken` ([`apps/web/src/utils/api-fetch.ts`](apps/web/src/utils/api-fetch.ts)). In other words, the PR claims "Token available for authenticated API calls", but no authenticated fetch path consumes `agentSessionToken` / `__babylonEmbedToken`, so the Milady credential pass-through does not actually work end-to-end.

2. `isEmbedded` currently flips on for `window.self !== window.top`, regardless of host or explicit opt-in ([`apps/web/src/contexts/EmbedContext.tsx`](apps/web/src/contexts/EmbedContext.tsx)). That means this Milady-specific flow will also activate inside every other iframe context we already support or may encounter: Discord Activity, Farcaster/Telegram wrappers, third-party previews, etc. Besides hiding app chrome unexpectedly, it also starts posting `BABYLON_READY` to any parent window. This needs a narrower detection rule, not a generic "any iframe" heuristic.

3. The incoming `BABYLON_AUTH` message is accepted from any origin ([`apps/web/src/contexts/EmbedContext.tsx`](apps/web/src/contexts/EmbedContext.tsx)). Right now any parent that can iframe Babylon can send a forged token/agentId pair and the app will trust it. Before merging, this needs origin validation against an explicit allowlist or another signed handshake mechanism.

Once those are addressed, I think the UI changes themselves are straightforward.

PR #1464, opened 2026-04-06
feat: add trading leaderboard metric

slkzgm:
## Summary
- add a canonical `metric=reputation|trading` axis to the leaderboard API, cache keys, and React Query hooks
- introduce `TradingLeaderboardService` for per-wallet and team trading rankings based on realized `lifetimePnL`
- update the leaderboard UI to support two axes: `Reputation|Trading` and `Per Wallet|Team`
- remove legacy leaderboard query params from the canonical contract and keep the trading leaderboard behavior explicit

## Trading ranking rules
- wallet trading ranks individual users/agents by realized `lifetimePnL` descending
- team trading ranks humans by `user.lifetimePnL + SUM(agent.lifetimePnL)`
- tie-breaks stay stable: score DESC, createdAt ASC, id ASC

## Validation
- `bunx biome check` on touched files
- `bun test packages/testing/unit/leaderboard-route.test.ts packages/testing/unit/leaderboard-cache.test.ts packages/testing/unit/leaderboard-query-hooks.test.ts packages/testing/unit/leaderboard-fetch-retry.test.ts apps/web/src/app/api/leaderboard/query.test.ts apps/web/src/app/api/leaderboard/__tests__/route.test.ts`

## Notes
- trading leaderboard currently includes wallets/teams at `0` lifetime PnL; filtering inactive/non-trading accounts is an intentional follow-up product decision, not part of this PR


SLKReview (changes requested):
Thanks for pushing this through. I reviewed the PR description, the code changes, and the current PR conversation.

Requesting changes for two issues before merge:

1. `packages/api/src/services/trading-leaderboard-service.ts:47-53` and `packages/api/src/services/trading-leaderboard-service.ts:201-291`
The new trading leaderboard adds two hot paths that sort/rank on `User.lifetimePnL`, but the schema still only has the existing `reputationPoints` index. Both the page query and the rank query will fall back to full scans/sorts as the `User` table grows, which is exactly the kind of production path we should not merge unmeasured. Please either add the supporting index/migration for the trading sort key, or attach `EXPLAIN ANALYZE` evidence that the current plan is acceptable at expected cardinality.

2. `packages/shared/src/validation/schemas/common.ts:443-460`
`metric` and `type` silently coerce invalid values back to the defaults. For example, `/api/leaderboard?metric=tradng` will return the reputation wallet leaderboard with `200` instead of surfacing a client bug. Since `apps/web/src/app/api/leaderboard/query.ts` already treats unsupported params as a `400`, the canonical contract should also reject invalid enum values instead of normalizing them. Otherwise the new axis is easy to misuse and very hard to debug.

Aside from those two concerns, the service split, cache key changes, and UI/query-key wiring are coherent.

SLKReview (approved):
The follow-up addresses the two blocking issues from my previous review:

- `User_lifetimePnL_createdAt_id_idx` was added for the new trading wallet sort/rank path.
- `metric` and `type` now use strict enums, so invalid values fail validation instead of silently falling back.

I re-checked the updated diff and don't see any new blocking issues in the leaderboard implementation. Approving.

PR #1465, opened 2026-04-06
feat: switch trading leaderboard to capital-adjusted return

slkzgm:
## Summary
- switch the trading leaderboard from raw lifetime PnL to capital-adjusted return
- add a canonical `TradingPerformanceService` to compute wallet/team capital base, effective capital, and trading return
- update the trading leaderboard payload and UI to rank on return while still showing lifetime PnL as secondary context

## Trading return formula
- `tradingReturn = lifetimePnL / max(capitalBase, 1000)`
- wallet capital base is derived from canonical balance funding ledger entries
- team capital base excludes owner-agent internal funding to avoid double counting

## Validation
- `bun test packages/testing/unit/trading-performance-service.test.ts`
- `bun test packages/testing/unit/leaderboard-route.test.ts`
- `bun test packages/testing/unit/leaderboard-cache.test.ts`

## Notes
- local working tree still contains unrelated modified files from prior formatting churn and untracked local files; they are intentionally not part of this PR
- the main technical debt left after this PR is that some capital-base classification still relies on ledger heuristics/transaction descriptions and should be canonized before reintroducing trading balance transfers

### DIFF
--- a/apps/web/src/app/api/leaderboard/route.ts
+++ b/apps/web/src/app/api/leaderboard/route.ts
@@ -4,7 +4,8 @@
  * Supports two ranking axes:
  * - **metric**
  *   - `reputation`: general leaderboard based on reputation points
- *   - `trading`: trading leaderboard based on realized lifetime P&L
+ *   - `trading`: trading leaderboard based on realized return
+ *     (`lifetimePnL / max(capitalBase, 1000)`) with lifetime P&L included as context
  * - **type**
  *   - `wallet`: per-wallet ranking (users and agents as individuals)
  *   - `team`: user + their agents combined
--- a/apps/web/src/app/leaderboard/fetchLeaderboardData.ts
+++ b/apps/web/src/app/leaderboard/fetchLeaderboardData.ts
@@ -8,6 +8,9 @@ export interface LeaderboardUser {
   reputationPoints: number;
   balance: number;
   lifetimePnL: number;
+  capitalBase?: number;
+  effectiveCapitalBase?: number;
+  tradingReturn?: number;
   createdAt: Date;
   rank: number;
   isAgent?: boolean;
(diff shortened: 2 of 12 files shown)


SLKReview (changes requested):
I found one correctness issue that blocks approval.

`packages/api/src/services/trading-performance-service.ts` counts `owner_deposit` toward wallet capital, but it never subtracts `owner_withdraw`. For agent wallets that have capital returned to their owner, the denominator stays inflated forever, so their trading return is understated and the wallet leaderboard becomes incorrect. That also diverges from the existing `totalDeposited - totalWithdrawn` economic baseline used elsewhere in the codebase.

Please treat `owner_withdraw` as a capital reversal in the wallet calculation and add regression coverage for an agent deposit -> owner withdrawal flow.

SLKReview (changes requested):
The `owner_withdraw` fix itself looks correct, but I still see one blocking issue in the follow-up.

`packages/api/src/services/trading-performance-service.ts` now exports `getCapitalBaseContribution`, while the actual leaderboard queries still use separate SQL `CASE` expressions. That leaves us with two independent implementations of the capital classification rules, and the new unit tests only exercise the helper, not the query path that actually ranks wallets and teams.

This adds a production-facing abstraction with no runtime callers, broadens the `@babylon/api` surface, and creates drift-prone duplication in exactly the part of the feature we were trying to lock down. Please either make the SQL builders derive from the same source of truth, or drop the helper and add coverage around the real service/query behavior instead.

SLKReview (approved):
The follow-up resolves the previous blocking issues.

`owner_withdraw` is now treated as a wallet capital reversal, and the capital classification used by the SQL queries is derived from a single shared rule table instead of a separate exported helper. That removes the drift risk in the earlier revision and keeps the production path as the source of truth.

I do not see any remaining code-level blockers in this PR.

PR #1466, opened 2026-04-06
refactor: canonize trading capital base classification

slkzgm:
## Summary
- introduce a canonical balance-transaction capital-base classification helper in `packages/db`
- refactor `TradingPerformanceService` to consume centralized wallet/team capital-base rules instead of inline heuristics
- reuse canonical balance description constants for daily login and agent registration refunds so capital-base exclusions are explicit and shared
- add focused unit coverage for classification behavior and trading performance floor behavior

## Notes
- wallet/team capital-base treatment is now explicit per balance transaction type
- the only remaining description-based logic is the intentionally bounded compatibility handling for legacy `deposit` rows, now centralized in one helper
- unrelated modified files remain in the local working tree due to prior formatting churn and are intentionally not part of this PR

## Validation
- `bun test packages/testing/unit/balance-transaction-classification.test.ts packages/testing/unit/trading-performance-service.test.ts`

### DIFF
--- a/packages/api/src/services/agent-evm-registration-service.ts
+++ b/packages/api/src/services/agent-evm-registration-service.ts
@@ -1,4 +1,12 @@
-import { and, balanceTransactions, db, eq, sql, users } from '@babylon/db';
+import {
+  AGENT_EVM_REGISTRATION_REFUND_BALANCE_DESCRIPTION,
+  and,
+  balanceTransactions,
+  db,
+  eq,
+  sql,
+  users,
+} from '@babylon/db';
 import {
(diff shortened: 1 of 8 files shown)


SLKReview (approved):
Applied a small follow-up in `d7e8d343` to keep the new TypeScript capital-base helper aligned with the SQL path for non-positive legacy `deposit` rows.

With that in place, the refactor matches the PR goal: the capital-base rules are centralized, the remaining description-based compatibility cases are explicit and shared, and the targeted unit coverage is in place.

PR #1468, opened 2026-04-06
chore: reconcile production into staging before release

slkzgm:
Summary:
- reconcile production history back into staging before the production release
- keep the current staging implementations for markets/schema/onchain cleanup
- include the release validation fixes needed for clean install, typecheck, unit tests, and build on staging

Validated:
- bunx biome check .
- bun run typecheck
- bun run test:unit
- bun run build

Notes:
- this PR restores a sane base so the follow-up staging to production PR stays small and predictable
- legacy market/onchain files from production were intentionally not reintroduced where staging has already removed them


SLKReview (changes requested):
I went through the reconciliation diff, the PR metadata, and the current conversation history. The schema-alignment fixes and the `netPeerTransfers` reconciliation look directionally right, but I don't think this is ready to approve yet.

1. `packages/testing/unit/wallet-actions-validation.test.ts:195` and `packages/testing/unit/wallet-actions-validation.test.ts:270`
The selector assertions were weakened from exact method IDs (`0xa9059cbb` / `0x42842e0e`) to generic `0x` / length checks. Nothing in the runtime path changed here, so this is pure coverage loss: the tests no longer prove that we are encoding the intended ERC-20 / ERC-721 entrypoints before sending calldata onchain. Please keep the selector-level assertions, or replace them with something equally specific.

2. `packages/testing/unit/auth/siwe.test.ts:186`
The expired-message case now uses an all-zero signature instead of a valid signed SIWE payload. That means the test only passes as long as `siweMessage.verify()` reports expiration before it reports an invalid signature. In other words, we stopped testing the real request path and started depending on verifier internals. Please switch this back to a deterministic valid signature fixture (or another real signed message) so the test still isolates the expiration branch without weakening signature coverage.

3. `apps/web/src/app/markets/_components/terminal/MarketsTradingTerminal.tsx:271`
This introduces a third `formatCompactNumber` variant, while the new comment says it is aligned with `packages/shared` T/Q support. It is not: `packages/shared/src/utils/format.ts:204` still only handles K/M, and `packages/testing/unit/shared/format.test.ts:103` still asserts that behavior. That leaves compact-number rules inconsistent across surfaces and adds another local formatter to maintain. Either promote the T/Q behavior into the shared helper, or move the markets-specific variant into the existing markets `_lib` formatter layer instead of duplicating it inside the terminal component.

slkzgm:
Addressed the requested changes:
- restored specific ERC-20 / ERC-721 entrypoint coverage by asserting the exact ABI signatures in the wallet calldata tests, while keeping the calldata structure checks
- replaced the expired SIWE test with a real signed expired payload path (deterministic signer selection + real signature) instead of the zero-signature shortcut
- promoted T/Q compact-number support into the shared formatter and removed the duplicate local formatter from MarketsTradingTerminal so the behavior is consistent across surfaces

Revalidated on the branch:
- bunx biome check .
- bun run typecheck
- bun run test:unit

SLKReview (changes requested):
Thanks for following up on the earlier review. The previous blockers are addressed, but I found one new regression in the shared formatter change.

1. `packages/shared/src/utils/format.ts:204`
Promoting the terminal formatter into `@babylon/shared` also changed the behavior for every value below `1000`: the old helper returned `num.toString()`, while the new version returns `Math.round(abs).toLocaleString()`. That means non-integer values now lose precision across all existing `formatNumber()` call sites, not just in markets. For example, `apps/web/src/components/admin/StatsTab.tsx:515` / `:523` / `:531` render per-tick averages through `formatNumber(...)`, so a value like `1.6` now becomes `2` instead of `1.6`.

I think the T/Q/non-finite support should land, but without changing the sub-1000 semantics for existing shared consumers. Preserving the previous `< 1000` branch (or adding a markets-specific formatter if the rounding is only desired there) would avoid that UI regression.

slkzgm:
Addressed the latest formatter review.

Change made:
- restored the historical behavior for compact numbers with absolute value below 1000 by returning num.toString() again
- kept the non-finite guard and the B/T/Q tiers for larger values
- added explicit test coverage for fractional and negative sub-1000 values

Validated on the PR branch:
- biome check on the touched formatter/test files
- shared format unit test passes
- typecheck passes

SLKReview (dismissed):
The remaining formatter regression is fixed. Restoring the `< 1000` branch to `num.toString()` keeps the new T/Q and non-finite handling for markets without changing existing shared consumers that rely on fractional values.

The added test coverage for fractional and negative sub-1000 inputs closes the gap as well. No further findings from my side on this revision.

PR #1469, opened 2026-04-07
fix: expose next at monorepo root for vercel

slkzgm:
Summary:
- add Next.js at the monorepo root package metadata so Vercel can detect the framework when rootDirectory is the repo root
- keep the fix minimal: package.json + bun.lock only

Why:
- the failed production deployment logs show Vercel finishing install, then erroring with: No Next.js version detected
- this repo keeps next in apps/web/package.json, but the Vercel project is configured with rootDirectory = null and framework = nextjs
- exposing the same Next.js version at the root fixes framework detection without changing runtime code

Validated:
- hotfix is intentionally metadata-only
- no app/runtime logic changed


slkzgm:
Retargeting this hotfix through staging first to keep the release flow clean and avoid reintroducing branch divergence.

PR #1479, opened 2026-04-07
fix: resolve prod 500s from cached bigint serialization

slkzgm:
## Summary
- make cache serialization/deserialization bigint-safe so identifier/user caches stop throwing before responses are built
- normalize leaderboard date/count fields when rows come back from SQL/cache as strings
- add regression tests for cache bigint round-tripping and cached ISO date handling in leaderboard ranking

## Prod errors addressed
- /api/users/[userId]/profile
- /api/users/[userId]/balance
- /api/users/[userId]/portfolio-breakdown
- /api/markets/positions/[userId]
- /api/agents/team-dashboard
- /api/leaderboard
- /api/leaderboard/me
- /api/users/[userId]/pnl-history is expected to benefit too because it uses the same cached identifier lookup path

## Validation
- bun run check
- bun run typecheck
- bun run lint
- bun run test:unit
- bun test packages/testing/unit/cache-service-serialization.test.ts packages/testing/unit/trading-performance-date-normalization.test.ts
- bun test packages/testing/unit/leaderboard-route.test.ts packages/testing/unit/leaderboard-cache.test.ts
- bun test apps/web/src/app/api/agents/team-dashboard/route.test.ts

## Notes
- local bun run build reaches the Next.js page-data phase and then fails on an existing environment/config issue unrelated to this patch: base identity registry address is not configured while collecting /api/auth/twitter/scopes/callback
- Vercel prod logs before this patch showed the same Do not know how to serialize a BigInt error across balance, portfolio-breakdown, markets/positions, team-dashboard and leaderboard routes

### DIFF
--- a/packages/api/src/cache/cache-service.ts
+++ b/packages/api/src/cache/cache-service.ts
@@ -16,6 +16,37 @@
 import { logger } from '@babylon/shared';
 import { getRedisClient, isRedisAvailable } from '../redis';
 
+const CACHE_BIGINT_MARKER = '__babylonCacheBigInt__';
+
+function cacheJsonReplacer(_key: string, value: unknown): unknown {
+  if (typeof value === 'bigint') {
+    return { [CACHE_BIGINT_MARKER]: value.toString() };
+  }
+  return value;
+}
+
+function cacheJsonReviver(_key: string, value: unknown): unknown {
+  if (
+    typeof value === 'object' &&
+    value !== null &&
+    CACHE_BIGINT_MARKER in value
+  ) {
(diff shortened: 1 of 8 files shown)


SLKReview (approved):
This patch is scoped well and addresses the reported failure modes without adding unnecessary abstraction.

The cache-service change restores shape parity between cache misses and cache hits for `bigint` fields, and the leaderboard normalization keeps cached SQL rows compatible with the existing ranking logic. I did not find correctness, DRY, or performance issues in the implementation.

PR #1484, opened 2026-04-09
fix: harden perp close settlement

slkzgm:
## Summary\n- serialize perp close settlement on the position row before applying state changes\n- reject dust partial closes\n- keep closed rows immutable at the adapter level\n\n## Context\nProduction incident on SPCX showed repeated close/partial-close credits on the same open position. This forward-port keeps the staging line aligned with the prod hotfix so the issue is not reintroduced on the next release.


revlentless:
LGTM!

PR #1486, opened 2026-04-09
feat: route all inference through ElizaCloud when ELIZACLOUD_API_KEY is set

revlentless:
**Summary:**

When `ELIZACLOUD_API_KEY` is set, all LLM and embedding inference in Babylon routes through ElizaCloud's OpenAI-compatible proxy instead of calling Groq, OpenAI, or Anthropic directly. Individual provider keys remain as fallbacks so nothing breaks without an ElizaCloud key.

**Files changed (6):**

| File | Change |
|---|---|
| `packages/engine/src/llm/openai-client.ts` | ElizaCloud added as Priority #1 in `BabylonLLMClient`; new `resolveElizaCloudConfig()` helper; `'elizacloud'` provider type |
| `packages/engine/src/llm/embedding-client.ts` | Prefers `ELIZACLOUD_API_KEY` → `ELIZACLOUD_API_URL/openai/v1`; falls back to `OPENAI_API_KEY` |
| `packages/agents/src/llm/direct-groq.ts` | `resolveGroqBaseURL()` returns ElizaCloud URL when configured; `callGroqDirect()` accepts `ELIZACLOUD_API_KEY` as fallback |
| `packages/agents/src/runtime/AgentRuntimeManager.ts` | `getModelSettings()` + `character.settings` include ElizaCloud vars; `groqPlugin`/`openaiPlugin` activate on `ELIZACLOUD_API_KEY` |
| `packages/api/src/services/claude-service.ts` | Routes Anthropic SDK to `ELIZACLOUD_API_URL/anthropic/v1` when ElizaCloud key is set |
| `.env.example` | New "ElizaCloud Inference" section; individual provider keys marked as optional fallbacks |

**Test plan:**
- [ ] Set `ELIZACLOUD_API_KEY=<key>` and `ELIZACLOUD_API_URL=<url>` in `.env`, unset all other provider keys
- [ ] Run `bun run dev` — confirm no "No API key configured" warnings
- [ ] Trigger a game tick — confirm `BabylonLLMClient` logs `Using ElizaCloud (unified inference)`
- [ ] Confirm `callGroqDirect` resolves to ElizaCloud base URL in agent logs
- [ ] Without `ELIZACLOUD_API_KEY`, confirm fallback to `GROQ_API_KEY` / `ANTHROPIC_API_KEY` still works

PR #1485, opened 2026-04-09
feat: replace Privy with Steward auth + remove crypto/blockchain stack

revlentless:
CURRENTLY DRAFT BECAUSE: 
- working on some ui improvements for the components that replacement the privy ui components
- want to get related PRs on steward and elizacloud merged first

**Summary:**

This PR completes two major tasks: strips all crypto/blockchain dependencies from Babylon and replaces Privy with Steward, a self-hostable open-source auth provider.

**What was removed:**
- All Solana, EVM, Anchor, Agent0, and `@privy-io/*` packages and code (~37k lines deleted)
- Wallet-based login flows, NFT gating hooks, on-chain registration (`8004`)
- `@babylon/contracts` package and all related `tsconfig`/`next.config` references

**What was added:**
- Steward service in `docker-compose.yml` with health check, DB provisioning script, and `steward:init` npm script
- `stewardId` column on the `User` table (migration `0067_add_steward_id.sql`) alongside `privyId` for migration continuity
- Backend auth middleware rewritten to verify Steward JWTs (`jose`)
- New API routes: `/api/auth/session` (cookie bridge), `/api/auth/farcaster`, `/api/auth/farcaster-miniapp`, `/api/auth/telegram-miniapp`
- Frontend `StewardAuthProvider`, rewritten `useAuth` hook, and `LoginModal` with Google, Discord, X, email/passkey via Steward SDK (`@stwd/sdk`)
- `.env.example` updated with all Steward variables (`STEWARD_API_URL`, `STEWARD_JWT_SECRET`, `STEWARD_TENANT_ID`, `STEWARD_TENANT_API_KEY`, `STEWARD_APP_URL`)

**Test plan:**
- [ ] `bun run steward:init` provisions tenant and prints `STEWARD_TENANT_API_KEY`
- [ ] `docker compose up steward` becomes healthy
- [ ] Login modal renders with OAuth buttons (Google/Discord/X) enabled immediately
- [ ] Email magic-link flow delivers email and redirects back to `/auth/callback`
- [ ] Farcaster and Telegram miniapp auth routes return 200 with valid JWT
- [ ] Existing users with `privyId` are not broken (column still present)

PR #1483, opened 2026-04-08
feat: remove crypto/blockchain stack — Phase 1 (Agent0, Solana, EVM, ERC-8004)

revlentless:
## Summary

- Strips all external blockchain dependencies from the codebase: Agent0/ERC-8004, Solana registry, EVM smart contracts, SIWE wallet login, Anvil, and associated infrastructure — ~30k lines deleted across 236 files.
- Replaces wallet-based auth surface with a clean internal-auth-only model; Privy email/social login remains intact for Phase 2 (Steward swap).
- Removes all crypto fields from the auth layer (`walletAddress`, `onChainRegistered`, `nftTokenId`, `agent0TokenId`) and gates disabled NFT page routes via middleware redirect.

## Type

- [x] Refactor / cleanup (no behavior change)

## Context / Links

- Phase 1 of a two-phase auth modernisation: remove crypto → replace Privy with Steward.
- Phase 2 (Steward) is blocked pending SDK/docs from the author.

## Scope (keep it focused)

- [x] This PR is focused on a single change/theme (not a catch‑all)
- [x] Drive‑by refactors are excluded or split into a separate PR
- [x] Non‑goals / follow‑ups are listed below (with links)

**Areas touched**
- [x] Web UI (`apps/web`)
- [x] Web API routes / SSE / A2A (`apps/web`)
- [x] Domain / game engine (`packages/engine`, `packages/core/*`)
- [x] Agents / runtime / A2A / MCP (`packages/agents`, `packages/a2a`, `packages/mcp`)
- [x] API infra (`packages/api`)
- [x] Shared types/utils (`packages/shared`)
- [x] Tests (`packages/testing`)

## Non-goals / follow-ups

- Phase 2: Replace Privy with Steward (blocked on Steward SDK/docs).
- DB cleanup: remove deprecated columns (`walletAddress`, `onChainRegistered`, `nftTokenId`, `agent0TokenId`, Solana fields) in a future migration once Steward is live.
- NFT code: kept but fully disabled via middleware + `NEXT_PUBLIC_NFT_ENABLED` flag. Delete entirely once NFT strategy is decided.

## Changes

**Deleted entirely:**
- `packages/contracts/` — all Solidity contracts and deployment tooling
- `packages/shared/src/contracts/`, `constants/chains.ts`, `constants/identity.ts`
- `packages/agents/src/{solana-registry,agent0,identity}/` — all on-chain agent identity
[...]

PR #1487, opened 2026-04-10
feat: Phase 2 — remove crypto stack, Steward auth, ElizaCloud inference

revlentless:
## Summary

Consolidates PRs #1483, #1485, and #1486 into a single clean PR. Closes those three.

### Phase 1 — Remove crypto/blockchain stack
- Remove Agent0 SDK, Solana, EVM, ERC-8004 registration
- Strip wallet-based auth and on-chain registration flows
- Remove `@babylon/contracts` and dead crypto references from tsconfig/next.config

### Phase 2 — Replace Privy with Steward auth
- Add Steward to Docker Compose (local dev) via sibling repo
- New `stewardId` column on `User` table (migration 0067)
- Rewrite backend auth middleware to verify Steward JWTs (cookie + Bearer fallback)
- Rewrite frontend auth: `StewardAuthProvider`, `useAuth`, `LoginModal` (email + OAuth)
- Auth callback pages for email magic link, OAuth, Farcaster, Telegram
- `bun run steward:init` script to provision tenant on first boot
- `scripts/migrate-privy-to-steward.ts` migration script (batched, concurrent, cached)

### Phase 3 — Route all inference through ElizaCloud
- `BabylonLLMClient` prefers `ELIZACLOUD_API_KEY` over individual provider keys
- `EmbeddingClient` prefers ElizaCloud
- `direct-groq.ts` and `claude-service.ts` proxy through ElizaCloud
- `AgentRuntimeManager` loads groq/openai plugins when `ELIZACLOUD_API_KEY` is set
- Falls back gracefully to `GROQ_API_KEY` / `OPENAI_API_KEY` / `ANTHROPIC_API_KEY`

### Bug fixes (caught during validation)
- Fix `/api/users/me` to look up users by primary key (`users.id`) instead of incorrectly using the Babylon snowflake ID as a `stewardId` filter — caused new Steward users to never be found → silent error → onboarding infinite loop
- Fix onboarding redirect race: expose `isOnboardingResolved` from `OnboardingProvider` context so the redirect effect waits for `isReadyToShow` to settle (was causing a one-tick race between Zustand store update and React local state)
- Fix `ProfilePageClient` rendering `profileDescription` (image-gen prompt) as visible bio text — now correctly shows `description`

## Test plan
- [ ] `bun run check` — Biome clean
- [ ] `bun run typecheck` — zero errors
- [ ] `bun run test:unit` — all pass
- [ ] Magic link login → profile setup → feed (no redirect loop)
- [ ] Google/Discord/Twitter OAuth buttons visible and enabled in login modal
- [ ] `ELIZACLOUD_API_KEY` set → LLM calls route through ElizaCloud proxy
- [ ] `ELIZACLOUD_API_KEY` unset → falls back to individual provider keys

Closes #1483
Closes #1485
Closes #1486


SLKReview (changes requested):
I found one blocking issue before merge:

1. `.env.example` now declares several keys twice (`ANTHROPIC_API_KEY`, `SIMULATION_BRIDGE_HOST`, `SIMULATION_BRIDGE_TOKEN`, `RPC_URL`, `SEPOLIA_RPC_URL`, `USE_MAINNET`). A few of them even conflict (`SIMULATION_BRIDGE_HOST=` earlier in the file vs `SIMULATION_BRIDGE_HOST=0.0.0.0` in the new legacy section). That makes the example env ambiguous, and any dotenv-style loader will apply last-one-wins semantics, so the earlier sections are no longer trustworthy as the documented defaults. Please keep a single declaration per variable and move the legacy notes into comments instead of re-declaring the same keys.

Aside from that, the Steward helper extraction is a good cleanup, the auth routes stay consistent, and the dead Privy/wallet exports are removed cleanly.

SLKReview (dismissed):
Follow-up cleanup commit applied to remove the duplicated `.env.example` declarations. I rechecked the current diff and I don't see any remaining blockers in this patch.

SLKReview (changes requested):
I found one blocker before merge:

1. The last ElizaCloud fix only updates `BabylonLLMClient` to use `${ELIZACLOUD_API_URL ?? 'https://elizacloud.ai'}/api/v1` plus `X-API-Key`, but the rest of the ElizaCloud call sites still point to `https://api.elizacloud.com` with the old `/openai/v1` or `/anthropic/v1` paths:
   - `packages/engine/src/llm/openai-client.ts:61-65,166-168,207-208`
   - `packages/engine/src/llm/embedding-client.ts:33-43`
   - `packages/agents/src/llm/direct-groq.ts:42-48`
   - `packages/api/src/services/claude-service.ts:27-34`
   - `.env.example:202`

That means turning on `ELIZACLOUD_API_KEY` can send different parts of the app to different proxies depending on which inference path is exercised. It also makes the sample env misleading, because copying `.env.example` still configures the old host while `BabylonLLMClient` now defaults to a different one.

Please pick one canonical ElizaCloud base/protocol and update every ElizaCloud consumer plus `.env.example` together in this patch, or revert the `openai-client`-only change.

PR #1491, opened 2026-04-10
feat: Feed QA — DB-backed dedup, cross-NPC similarity, chat quality guards

revlentless:
## Summary

**Mission**: QA Data feeds — make sure the data is good, make sure the feed NPC chat is good, make sure no duplicate posts, etc — feed should be good.

### Root causes fixed

1. **Arc event coverage was in-memory only** — `NewsArticlePacingEngine.arcEventCoverage` reset on every restart/cold start, causing duplicate articles for the same arc event after each deploy.

2. **BreakingArticleRateLimiter was in-memory** — breaking article timestamps were lost on restart, causing breaking-article bursts post-deploy.

3. **Cross-NPC topic dedup was per-process** — `TopicDiversityService` (singleton, per-agent) couldn't see what other NPCs posted in other processes. Two NPCs could post near-identical content in the same tick.

4. **Group chat had no message quality guards** — one-word fragments and near-duplicate messages within a chat window could be posted.

### Changes

| File | Change |
|------|--------|
| `packages/db/src/schema/narrative.ts` | New `arc_event_coverage` table |
| `packages/db/drizzle/migrations/0068_add_arc_event_coverage.sql` | DB migration |
| `packages/engine/src/services/event-generation-helpers.ts` | Replace `arcEventPacer` with DB reads/writes for arc coverage; `hasEventBeenCovered` and `markEventAsCovered` are now `async` DB-backed |
| `packages/engine/src/services/article-rate-limiter.ts` | `BreakingArticleRateLimiterService` now counts from DB (posts table) instead of in-memory timestamps; `tryReserveSlot`, `canGenerateArticle`, `getRemainingSlots` are now `async` |
| `packages/agents/src/autonomous/DirectExecutors.ts` | `executeDirectPost` queries last 20 recent posts, rejects if Jaccard similarity ≥ 0.5 with any NPC post in the past 10 min |
| `packages/engine/src/services/npc-group-dynamics-service.ts` | `postGroupMessages` rejects messages < 20 chars, < 3 words, or ≥ 0.5 Jaccard similarity to any message in the past 5 min in the same chat |
| `scripts/feed-health.ts` | New dev tool: `bun run report:feed` — post distribution, duplicate detection, article flood, group chat quality, feed staleness |
| `packages/testing/unit/feed-dedup.test.ts` | 24 new unit tests |
| `packages/testing/integration/feed-health.integration.test.ts` | 13 new integration tests |

### Dev tool

```bash
bun run report:feed           # human-readable with [OK]/[WARN] flags
bun run report:feed -- --json # machine-readable JSON
bun run report:feed -- --hours=6  # narrow window
```

## Test plan

- [x] `bun run check` — Biome format + lint, zero warnings
- [x] `bun run typecheck` — all packages pass
- [x] `bun run lint` — turbo lint, zero warnings
- [x] `bun test packages/testing/unit/feed-dedup.test.ts` — 24/24 pass
- [x] `bun test packages/testing/integration/feed-health.integration.test.ts` — 13/13 pass
- [ ] Run `bun run report:feed` against live DB after migration to verify health output
- [ ] Run `bun run db:migrate` to apply migration 0068

### DIFF
--- a/apps/web/src/app/api/cron/article-tick/route.ts
+++ b/apps/web/src/app/api/cron/article-tick/route.ts
@@ -412,9 +412,9 @@ export const POST = withErrorHandling(async function POST(_req: NextRequest) {
       const event = activeEventsData.activeEvents[eventIndex];
 
       if (event) {
-        // Check if we've already covered this event using explicit tracking
+        // Check if we've already covered this event using DB-backed tracking
         const eventId = event.questionId;
-        const alreadyCovered = hasEventBeenCovered(eventId);
+        const alreadyCovered = await hasEventBeenCovered(eventId);
 
         if (!alreadyCovered) {
           // Pick a random news org to write the article
@@ -432,8 +432,8 @@ export const POST = withErrorHandling(async function POST(_req: NextRequest) {
 
           if (result.status === 'success') {
             articlesCreated++;
-            // Mark this event as covered for future duplicate detection
-            markEventAsCovered(eventId, org.id, result.id);
+            // Mark this event as covered for future duplicate detection (DB-backed)
+            await markEventAsCovered(eventId, org.id, result.id);
             logger.info(
               `Article created by ${org.name}`,
               { eventId: event.questionId, articleId: result.id },
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "solana:registration:debug": "bun run scripts/debug-solana-registration.ts",
     "seed:nft-snapshot:csv": "bun run scripts/seed-nft-snapshot-csv.ts",
     "users:export:newsletter": "bun run scripts/export-newsletter-users-csv.ts",
+    "report:feed": "bun run scripts/feed-health.ts",
     "report:markets": "bun run scripts/market-diversity-report.ts",
     "report:markets:health": "bun run scripts/market-health.ts",
     "report:realism": "bun run scripts/market-realism-report.ts",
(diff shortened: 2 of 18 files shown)


SLKReview (changes requested):
Thanks for pushing this forward. Moving the feed-quality state out of process memory is the right direction, but I don't think this is merge-ready yet.

1. The new breaking limiter still does not enforce the ceiling it documents.
`BreakingArticleRateLimiterService.tryReserveSlot()` only reserves a single slot, but `maybeGenerateBreakingArticle()` immediately calls `generateArticlesForArcEvent(..., { skipRateLimit: true })`, and that path can still publish up to two org articles for the same event. With the new default of `DEFAULT_MAX_ARTICLES_PER_HOUR + 1 = 3`, the code path still allows `0 regular + 2 breaking + 2 breaking = 4` articles inside one hour: the second breaking event is accepted when the DB count is 2, reserves one slot, then emits two more articles. That means the comment at `article-rate-limiter.ts` is no longer true, and the post-deploy burst the PR is trying to eliminate is still possible. Please either reserve the real number of slots that can be consumed, or cap the breaking path to one article when this total-article budget is used.

2. The checked-in unit suite for `BreakingArticleRateLimiterService` was not updated to the new async / DB-backed contract.
`packages/engine/src/__tests__/article-rate-limiter.test.ts` still asserts the old default (`1`), still calls `getRecentArticleCount()` / `tryReserveSlot()` synchronously, and still expects `recordBreakingArticle()` to mutate local state. `scripts/test-unit-isolated.ts` collects tests from `packages/engine/src`, so this file is still part of `bun run test:unit`. Until that suite is rewritten (or intentionally excluded), the test plan in the PR description is out of sync with what the repository actually checks.

Once those two issues are addressed, I don't see anything else blocking from the code I reviewed.

PR #1500, opened 2026-04-17
fix(auth): provision steward users for farcaster miniapp sign-ins

0xSolace:
## Summary
- fix the miniapp Farcaster auth route so Steward provisioning runs on every sign-in, even when no real email is present
- use the same placeholder email convention as the main Farcaster route: `fid-{FID}@farcaster.babylon.local`
- include a Farcaster-derived placeholder name using the quickAuth custody address when present

## Bug
This route had the same ghost-user bug addressed in #1499 on `/api/auth/farcaster/route.ts`: the API minted a Babylon Steward JWT, but `ensureStewardUser()` only ran when an email existed. For Farcaster miniapp logins, that meant users could receive a JWT without a real Steward user record behind it.

## Notes
- quickAuth payload shape differs from SIWF: it exposes `sub` for FID and `address` for custody address, so this route derives placeholder identity from those fields instead of parsing the SIWF message
- this PR is a standalone follow-up and does not depend on #1499, because it branches from `origin/staging` and inlines the name-aware provisioning logic needed here
- placeholder email convention intentionally matches the main Farcaster route for consistency

### DIFF
--- a/apps/web/src/app/api/auth/farcaster-miniapp/route.ts
+++ b/apps/web/src/app/api/auth/farcaster-miniapp/route.ts
@@ -22,14 +22,24 @@ import { createClient } from '@farcaster/quick-auth';
 import { SignJWT } from 'jose';
 import { NextRequest, NextResponse } from 'next/server';
 
+import { getStewardJwtSecret } from '@/lib/auth/steward-server';
+
 const STEWARD_API_URL = process.env.STEWARD_API_URL ?? 'http://localhost:3200';
 const STEWARD_PLATFORM_KEY =
   (process.env.STEWARD_PLATFORM_KEYS ?? '').split(',')[0]?.trim() ?? '';
-const STEWARD_JWT_SECRET = new TextEncoder().encode(
-  process.env.STEWARD_JWT_SECRET ?? 'dev-jwt-secret-change-in-prod'
-);
 
-async function ensureStewardUser(email?: string): Promise<string> {
+interface EnsureStewardUserOptions {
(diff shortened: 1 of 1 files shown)


SLKReview (changes requested):
Thanks for chasing the miniapp side of the ghost-user bug. The route is moving in the right direction, but I’m requesting changes for two blocking issues:

1. `apps/web/src/app/api/auth/farcaster-miniapp/route.ts:167` inserts `name: placeholderName` into `users`. The current `User` schema does not have a `name` column, so this write is out of sync with the rest of the codebase. This should go to an existing user field such as `displayName`, or be omitted entirely.

2. `apps/web/src/app/api/auth/farcaster-miniapp/route.ts:31` reimplements `ensureStewardUser()` locally even though we already extracted the shared Steward auth helpers into `apps/web/src/lib/auth/steward-server.ts` for this exact reason. That reintroduces duplication in the auth layer and makes the miniapp route drift from the shared provisioning path. The cleaner fix is to extend the shared helper to accept the optional placeholder `name`, then reuse it here.

I prepared a minimal patch for those two points here in case it helps:
https://github.com/BabylonSocial/babylon/commit/f8166d1c4a900a7acf1c821da35fa01577e848a5

Residual risk: there is still no targeted coverage for the quickAuth path, so once the route is adjusted I’d like at least one focused test around the email-less Farcaster sign-in flow.

0xSolace:
Superseded by #1499 (consolidated). Closing.

